PR #948, opened 2013-07-16
Provided a static lookAt method for Quaternion (Unity already does this)

dcerisano:
The following implementation of lookAt returns a Quaternion that is oriented with a forward and up vector!

Implemented to workaround problems with the existing BaseObject3D lookAt() method which disables it's own orientation quaternion and rotational matrix.

This uses Gram-Schmidt orthonormalization to form the required axes.

Tested and working beautifully (no gimbal lock, very smooth predictable behaviour).
Pretty sure this is how Unity does it, too.

To set a camera's orientation with this, make sure isCamera is true.
To set an object's orientation with this, make sure isCamera is false.

for example, to keep a player always oriented "up" just do this:
player.setOrientation(Quaternion.lookAt(forward, new Vector(0,1,0), false));

These can be placed directly into the Quaternion class as statics.

### DIFF
--- a/src/rajawali/math/Quaternion.java
+++ b/src/rajawali/math/Quaternion.java
@@ -577,4 +577,54 @@ public final class Quaternion {
             normalize();
         } 
     }
+    
+    
+	public static Quaternion lookAt(Vector3 lookAt, Vector3 upDirection, boolean isCamera) 
+	{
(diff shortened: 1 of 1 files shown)


dcerisano:
It would be nice if GLSL and GLES could be integrated with hardware accelerated quaternion operators. The math is more like calculus and much easier to parse than verbose matrix algebra commonly found in shaders.

PR #1129, opened 2013-11-10
Animation texture

ravenfeld:
Creating animation to play on texture.
- Make them appear texture
- Make them disappear texture
- Scroll a texture
...


ToxicBakery:
Guess its been over a year now and I doubt I'll ever get around to this. If you would like to make a clean single commit pull request instead of the manual merges like this one contained, I would be happy to accept the request.

PR #1347, opened 2015-02-04
Added Shield

ghost:
Added a shield.io badge for android-libs.com; a repository with hundreds of android libraries!


jwoolston:
I'm going to say @MasDennis is the only one who can decide on this.

PR #1349, opened 2015-02-10
Added Prism primitive.

mcteo:
Added a new type of primitive. If it works well, the Cube primitive can be rewrote as a special case of this. Can someone double check the vertices and normals arrays?


ToxicBakery:
Looks fine to me, thanks

PR #1351, opened 2015-03-05
Fixed a bug in the equals function

Artimunor:
Fixed a bug in the equals function that caused it to return always false
if dot(other) returned > 1.0
Thanks to @jwoolston for providing the fix.


jwoolston:
Looks good to me.

PR #1353, opened 2015-03-06
fixed clone rotation and scale cloning

Artimunor:
When cloning an Object3D the rotation and scale became pointers to the
original object, causing scaling or rotating an object to influence the
source and all other clones of that source object

### DIFF
--- a/src/rajawali/Object3D.java
+++ b/src/rajawali/Object3D.java
@@ -649,8 +649,8 @@ public class Object3D extends ATransformable3D implements Comparable<Object3D>, 
 	public Object3D clone(boolean copyMaterial, boolean cloneChildren) {
 		Object3D clone = new Object3D();
 		cloneTo(clone, copyMaterial);
-		clone.setRotation(getRotation());
-		clone.setScale(getScale());
+        clone.setRotation(getRotation().clone());
+        clone.setScale(getScale().clone());
 		
 		if(cloneChildren)
 		{


src/rajawali/Object3D.java
```diff
@@ -649,8 +649,8 @@ protected void cloneTo(Object3D clone, boolean copyMaterial) {
 	public Object3D clone(boolean copyMaterial, boolean cloneChildren) {
 		Object3D clone = new Object3D();
 		cloneTo(clone, copyMaterial);
-		clone.setRotation(getRotation());
-		clone.setScale(getScale());
+        clone.setRotation(getRotation().clone());
```
ToxicBakery: This is redundant as the setRotation method calls setAll on the vector which copies the internal double values.

https://github.com/MasDennis/Rajawali/blob/master/src/rajawali/ATransformable3D.java#L212

Which then calls

https://github.com/MasDennis/Rajawali/blob/master/src/rajawali/math/vector/Vector3.java#L167

src/rajawali/Object3D.java
```diff
@@ -649,8 +649,8 @@ protected void cloneTo(Object3D clone, boolean copyMaterial) {
...
+        clone.setScale(getScale().clone());
```
ToxicBakery: This seems to be the same as above, setAll is used so each Vector on each object should be unique.

src/rajawali/Object3D.java
Artimunor: hi there
i can assure you it is bugged
that was the reason i went in and fixed it
when i scaled a model, all clones of it scaled also

PR #1358, opened 2015-03-09
Manager initialization

Artimunor:
moved initialization of TextureManager and MaterialManager from
onSurfaceChanged to onSurfaceCreated so Textures and Materials can be
loaded in an override of this method (like google strongly advises in
their developer presentations)


jwoolston:
I have tested this and can't find any issues with it, and from a conceptual standpoint, it makes sense to me. Additionally, it will be easy to reverse if need be.

PR #1393, opened 2015-03-19
Async parsing

### LINKED ISSUE #370: Make parsers thread safe
### Issue
It seems trying to parse in threads does not work properly. When this library was created devices mostly had only 1 core processors and it would have been pointless to try and thread something that is going to be maxing that core. Now that most devices are multi-core it may be useful to make model loading thread friendly. 

### Pros
* Potentially faster model loading
* A model manager for managing loading could permit additional optimizations to aid users.

### Cons
* Someone would have to do the work and we already have a pretty big back log
* Loading from the file system is expensive and multiple threads does not necessarily mean a speed gain would actually occur.
* This would add to the complexity of the model parsers which seem to already have enough problems without throwing more complexities at them.

Anyone have any thoughts on this?

jwoolston:
Fixes #370. Adds async loading the `RajawaliScene`. User code can register a callback to be notified when loading has completed or failed. They may then do whatever they wish with this information. The typical use case would be to add a parsed model to the scene.

I did also rename the `parsers` package to `loader` to match the rest of the naming convention.

**CAVEAT:** Just because its happening on another thread doesnt mean the load cant impact the render.


jwoolston:
@ToxicBakery let me know what you think.

jwoolston:
Merge with https://github.com/Rajawali/RajawaliExamples/pull/49

rajawali/src/main/java/rajawali/RajawaliSupportFragment.java
```diff
@@ -86,8 +86,14 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container,
     @Override
     public void onDestroy() {
         super.onDestroy();
-        unbindDrawables(mLayout);
-        System.gc();
+        try {
+            mRenderer.onSurfaceDestroyed();
+            mRenderer = null;
+            unbindDrawables(mLayout);
+            System.gc();
```
ToxicBakery: superfluous

rajawali/src/main/java/rajawali/RajawaliSupportFragment.java
```diff
@@ -86,8 +86,14 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container,
     @Override
     public void onDestroy() {
         super.onDestroy();
-        unbindDrawables(mLayout);
-        System.gc();
+        try {
+            mRenderer.onSurfaceDestroyed();
+            mRenderer = null;
```
ToxicBakery: superfluous

rajawali/src/main/java/rajawali/loader/LoaderAWD.java
```diff
@@ -10,35 +10,35 @@
...
+import rajawali.loader.awd.BlockSharedMethod;
+import rajawali.loader.awd.BlockSimpleMaterial;
+import rajawali.loader.awd.BlockSkeleton;
+import rajawali.loader.awd.BlockSkeletonAnimation;
+import rajawali.loader.awd.BlockSkeletonPose;
+import rajawali.loader.awd.BlockSkybox;
+import rajawali.loader.awd.BlockTextureProjector;
+import rajawali.loader.awd.BlockTriangleGeometry;
+import rajawali.loader.awd.BlockUVAnimation;
+import rajawali.loader.awd.exceptions.NotImplementedParsingException;
```
ToxicBakery: ugh finally

rajawali/src/main/java/rajawali/renderer/RajawaliRenderer.java
```diff
@@ -132,20 +147,33 @@ public RajawaliRenderer(Context context) {
...
-		mSceneQueue = new LinkedList<AFrameTask>();
-		mSceneCachingEnabled = true;
-		mSceneInitialized = false;
+        mRenderTargets = Collections.synchronizedList(new CopyOnWriteArrayList<RenderTarget>());
+        mSceneQueue = new LinkedList<AFrameTask>();
+        mSceneCachingEnabled = true;
+        mSceneInitialized = false;
+
+        mLoaderThreads = new SparseArray<ModelRunnable>();
+        mLoaderCallbacks = new SparseArray<IAsyncLoaderCallback>();
```
ToxicBakery: Are the generic definitions required here? If not we should aim to start removing the J7 notices about it.

rajawali/src/main/java/rajawali/renderer/RajawaliRenderer.java
```diff
@@ -1310,4 +1410,63 @@ public void setUsesCoverageAaAll(boolean usesCoverageAa) {
 			}
 		}
 	}
+
+    @SuppressLint("HandlerLeak")
```
ToxicBakery: The suppress may be warranted here as items in the async queue can still be processing after the renderer is stopped as nothing seems to stop the queue. Unless I missed it, the queue of loaders should be nuked and the callbacks released.

rajawali/src/main/java/rajawali/renderer/RajawaliRenderer.java
```diff
@@ -1310,4 +1410,63 @@ public void setUsesCoverageAaAll(boolean usesCoverageAa) {
...
+
+        @Override
+        public void handleMessage(Message msg) {
+
+            final int id = msg.arg2;
+            final ALoader loader = mLoaderThreads.get(id).mLoader;
+            final IAsyncLoaderCallback callback = mLoaderCallbacks.get(id);
+
+            mLoaderThreads.remove(id);
+            mLoaderCallbacks.remove(id);
```
ToxicBakery: If the loader crashes, the callback will never be released. See the above comment.

rajawali/src/main/java/rajawali/scene/RajawaliScene.java
```diff
@@ -163,7 +174,7 @@ public RajawaliScene(RajawaliRenderer renderer, GRAPH_TYPE type) {
 	 * method. 
 	 */
 	protected void initSceneGraph() {
-		switch (mSceneGraphType) { //Contrived with only one type I know. For the future!
+		switch (mSceneGraphType) { //I know its contrived with only one type. For the future!
```
ToxicBakery: Well add more types already, slacker

rajawali/src/main/java/rajawali/renderer/RajawaliRenderer.java
jwoolston: @ToxicBakery the generic definitions are to support J6 devices.

PR #1400, opened 2015-03-21
Cloned Object3Ds now get names from their origin Object3D fixes #1399

### LINKED ISSUE #1399: Cloned Object3Ds don't have names
When an `Object3D` is cloned from another `Object3D`, the name attribute should be cloned too.

This is very problematic if you want to get Childs of a cloned `Object3D` as you can't use `getChildByName()` anymore.

C0br4:
fixes #1399

### DIFF
--- a/rajawali/src/main/java/rajawali/Object3D.java
+++ b/rajawali/src/main/java/rajawali/Object3D.java
@@ -1,11 +1,11 @@
 /**
  * Copyright 2013 Dennis Ippel
- * 
+ *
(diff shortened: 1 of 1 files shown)


jwoolston:
Once the Travis build completes I'll merge it. As a note, please refrain from auto-formatting files when sending pull requests as it makes it difficult to find changes. I know the formatting is a mess - we have no standard defined for that and @MasDennis and myself in particular have very different ideas of what it should be.

PR #1434, opened 2015-04-01
Debug visualizer

MasDennis:
Finally added this debug visualizer class. It can be used to draw a grid floor, visualize camera frustums and visualize lights. I will add a trident as well.
Camera frustum visualization was already there but I've removed it out of the scene and camera classes. This is a much cleaner solution.
I will also add an example to the examples app.


jwoolston:
@MasDennis This will be tremendously helpful following the `v1.0` release when I finish up the scene graph

Issue #1336, opened 2014-12-24
Live Wallpaper Memory Leak on Changing Textures on Runtime


abhishekBansal:
Ok, I have seen many previous posts about this here but none of them actually helped. I am building a LWP using Rajawali which requires change of texture on a quad at runtime when user changes prefs. After changing texture a few times it crashes with out of memory errors. Also at few times i see black screen in preview mode or in live mode. 

```
java.lang.OutOfMemoryError
            at android.graphics.BitmapFactory.nativeDecodeAsset(Native Method)
            at android.graphics.BitmapFactory.decodeStream(BitmapFactory.java:594)
            at android.graphics.BitmapFactory.decodeResourceStream(BitmapFactory.java:429)
            at android.graphics.BitmapFactory.decodeResource(BitmapFactory.java:452)
            at android.graphics.BitmapFactory.decodeResource(BitmapFactory.java:482)
            at rajawali.materials.textures.ASingleTexture.setResourceId(ASingleTexture.java:93)
            at rajawali.materials.textures.Texture.<init>(Texture.java:32)
            at com.rrapps.Renderer.initScene(InfiniteTunnelRenderer.java:55)
            at rajawali.renderer.RajawaliRenderer.onSurfaceChanged(RajawaliRenderer.java:426)
            at android.opengl.GLSurfaceView$GLThread.guardedRun(GLSurfaceView.java:1512)
            at android.opengl.GLSurfaceView$GLThread.run(GLSurfaceView.java:1240)
```

here is how i am changing texture when user changes shared preferences

```
private void _handleTextureChange() {
        int currentTextureNo = TApplication.SettingsInstance.getCurrentTextureNo();

        if(currentTextureNo == -1) {
            Log.e(LogTag, "Wrong texture Id(-1) from shared prefs");
        }

        // remove existing texture
        ATexture existingTexture;
        if(mMaterial != null) {
            existingTexture = mMaterial.getTextureList().get(0);
            mTextureManager.removeTexture(existingTexture);

            ATexture newTexture;
            int drawableId = 0;
            switch (currentTextureNo) {
                case 1:
                    drawableId = R.drawable.t1;
                    break;
                case 2:
                    drawableId = R.drawable.t2;
                    break;
                case 3:
                    drawableId = R.drawable.t3;
                    break;
                case 4:
                    drawableId = R.drawable.t4;
                    break;
                default:
                    drawableId = R.drawable.t2;
                    break;
            }

[...]

PR #1435, opened 2015-04-02
Fixes leaks with managers (#1336)

Jaben:
Right now renderers are being registered twice in both TextureManager and MaterialManager: on render constructor and in onSurfaceCreated. They are also re-added on subsequent calls to onSurfaceCreated.


jwoolston:
I have actually seen this come up intermittently in the last few days. I dont see an issue with the patch.

PR #1479, opened 2015-04-17
Bugfix: TypedArray.get for custom styles takes attribute index, not array index

gormanb:
RajawaliSurface currently uses the attribute counter as the input to TypedArray.getXYZ when parsing the custom surfaceview style, whereas it should be using the RajawaliSurface_[attribute] values.


jwoolston:
@gormanb before I merge, could you apply the same fix to `RajawaliTextureView`?

gormanb:
@jwoolston Yep, just realised that it would be in there too - will push in a sec.

jwoolston:
@gormanb Thanks for the fix!

gormanb:
@jwoolston No problem! :beer:

PR #1486, opened 2015-05-03
Fixes semantic issues with a couple of methods.

gormanb:
Vector3.normalize's contract stipulates that it will return the original magnitude of the vector; currently it returns the reciprocal of the magnitude. Similarly, the setRotation methods in ATransformable should set the absolute value of mOrientation from the supplied arguments, but currently apply the rotation to the existing orientation and are therefore indistinguishable from the rotate() methods.


ToxicBakery:
Looks good to me, @jwoolston ?

jwoolston:
Seems fine. I meant to merge earlier but github on the phone doesn't like you to be productive.

PR #1488, opened 2015-05-08
Pick a suitable config (8888) when background is transparent

comac2k:
RajawaliSurfaceView was initialized with 0 bits for mBitsAlpha and there was no way to alter that value programatically. Not every combination of RGB and A sizes works. So, I force 8888 when mIsTransparent.


jwoolston:
Good find.

PR #1490, opened 2015-05-15
Delay count wasn't set to 0 when reset() was called.

MasDennis:


ToxicBakery:
This was intentional, reset is only intended to halt the animation and rewind its elapsed animation time. You'll notice that none of the other fields are reset either like reverse, repeat count, etc. 

For clarity the controls are intended be timeline oriented with play, pause, stop, reset all intended to relate with time not features.

ToxicBakery:
Nvm my previous message, I should drink coffee to avoid looking like an idiot. Carry on.

MasDennis:
Or maybe a beer instead of a coffee? It is almost 5 o'clock in Armenia.

ToxicBakery:
beer isn't for another 5 hours when its 6 o'clock here.

PR #1508, opened 2015-06-10
Fixed screen quad texture coordinates.

MasDennis:


adamantivm:
May I ask you a quick question related to this?
As [per your suggestion](https://github.com/Rajawali/Rajawali/issues/1518#issuecomment-113962896) I was using ScreenQuads to render RajawaliScenes. 
After this PR though the scenes started rendering upside down. Is there anything I can do to the RenderTarget or the ScreenQuad to re-mirror it back so that up becomes up again?
The only alternative I can think of is to create my own ScreenQuad class as a copy of this one and setting the texture coordinates back to what they were, but it doesn't sound like the right solution.

PR #1524, opened 2015-06-25
Remove redundant code?

adamantivm:
This is a very minor PR, just trying to point out what I think is a small mistake?
Note that I am very very far from an expert but my common sense reading this code makes me feel the texture attachment GL calls are done twice. (once in the code block I removed and then again in the ```bindTextureByName``` method above)
If that duplication was intentional please disregard this PR.


MasDennis:
Thanks @adamantivm you're absolutely right. Some duplicate code there.

PR #1531, opened 2015-07-29
Fixed equals(Vector3 obj, double error)...

rpicolet:


ToxicBakery:
This looks like a good PR to me. @jwoolston ?

jwoolston:
Oops. My mistake. Good catch @rpicolet

PR #1544, opened 2015-08-26
"Picture-in-picture" renderer contribution

adamantivm:
This is a proposed contribution based on the [RajawaliSideBySideRenderer](https://github.com/Rajawali/Rajawali/blob/master/rajawali/src/main/java/org/rajawali3d/renderer/RajawaliSideBySideRenderer.java) which allows rendering two separate scenes in the same view / GLES thread.

[Here you can find](https://github.com/creativa77/RajawaliExamplePip) a very simple example use of this contribution.

Looking forward to comments and suggestions for improvement.

### DIFF
--- a/rajawali/build.gradle
+++ b/rajawali/build.gradle
@@ -29,17 +29,9 @@ android {
             ext."signing.password" = System.getenv('SIGNING_KEY_PASSWORD') != null ? System.getenv('SIGNING_KEY_PASSWORD') : System.properties['SIGNING_KEY_PASSWORD'] ? System.properties['SIGNING_KEY_PASSWORD'] : hasProperty('SIGNING_KEY_PASSWORD') ? property('SIGNING_KEY_PASSWORD') : ""
             ext."signing.secretKeyRingFile" = "./rajawali_secret.gpg"
 
-            // version = rootProject.getReleaseVersion(property('VERSION_NAME'))
+            version = rootProject.getReleaseVersion(property('VERSION_NAME'))
         }
     }
-}
-
-def isTag() {
-    return false
-}
-
-def getBranchName() {
-    return 'master'
 }
 
 dependencies {
--- a/rajawali/build.gradle
+++ b/rajawali/build.gradle
@@ -29,9 +29,17 @@ android {
             ext."signing.password" = System.getenv('SIGNING_KEY_PASSWORD') != null ? System.getenv('SIGNING_KEY_PASSWORD') : System.properties['SIGNING_KEY_PASSWORD'] ? System.properties['SIGNING_KEY_PASSWORD'] : hasProperty('SIGNING_KEY_PASSWORD') ? property('SIGNING_KEY_PASSWORD') : ""
             ext."signing.secretKeyRingFile" = "./rajawali_secret.gpg"
 
-            version = rootProject.getReleaseVersion(property('VERSION_NAME'))
+            // version = rootProject.getReleaseVersion(property('VERSION_NAME'))
         }
     }
+}
+
+def isTag() {
+    return false
+}
+
+def getBranchName() {
+    return 'master'
 }
 
 dependencies {
--- a/rajawali/src/main/java/org/rajawali3d/renderer/RajawaliPipRenderer.java
+++ b/rajawali/src/main/java/org/rajawali3d/renderer/RajawaliPipRenderer.java
@@ -48,11 +48,11 @@ public class RajawaliPipRenderer extends RajawaliRenderer {
 
     // These variables control where the minimap is placed. Note they are specified in standardized
     // OpenGL coordinates 0 to 1
-    private final float pipScale;
-    private final float pipMarginX;
-    private final float pipMarginY;
+    private final float mPipScale;
+    private final float mPipMarginX;
+    private final float mPipMarginY;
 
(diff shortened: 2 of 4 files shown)


ToxicBakery:
Oh man, a pull request that passes the build. I salute you sir. Review incoming.

jwoolston:
@adamantivm Nicely done. :clap:

adamantivm:
Hahahaha, thanks. I suppose the key is to put code in separate classes the build tests don't know about? :)

adamantivm:
Any chance to have this looked at, at least a first quick review pass for general feedback?

rajawali/src/main/java/org/rajawali3d/renderer/RajawaliPipRenderer.java
```diff
@@ -0,0 +1,192 @@
...
+        setupMiniTouchLimits();
+
+        mMiniQuad = new WorkaroundScreenQuad();
+        // Set the size of the mini view using a scale factor (pipScale times the main view)
+        mMiniQuad.setScale(pipScale);
+        // Position the mini view in the top right corner
+        // For X and Y, the position is:
+        //   50% screen shift to the right/top minus half the size of the minimap to bring it back
+        //   left/bottom into full view plus a little bit more left/bottom to leave margin
+        mMiniQuad.setX(.5 - pipScale / 2 - pipMarginX / (2 * mDefaultViewportWidth));
```
ToxicBakery: I would prefer mixing floats/doubles/integers as it can cause logic issues and confuse intent. Can you please update the .5 to .5d and specify the 2 instances as 2d unless of course this needs to all be float.

ToxicBakery:
If you fix the above comment I'm ok with everything else. @jwoolston anything you want changed?

rajawali/src/main/java/org/rajawali3d/renderer/RajawaliPipRenderer.java
```diff
@@ -0,0 +1,192 @@
...
+    private float miniXmin, miniXmax, miniYmin, miniYmax;
+
+    /**
+     * @param pipScale   Size of the mini view from 0 to 1, i.e.: ratio of full screen to mini
+     * @param pipMarginX Margin space from the mini view to the edge of the main view in pixels
+     * @param pipMarginY Margin space from the mini view to the edge of the main view in pixels
+     */
+    public RajawaliPipRenderer(Context context, float pipScale, float pipMarginX, float pipMarginY) {
+        super(context);
+        this.pipScale = pipScale;
```
jwoolston: Its kind of a personal preference thing, but I very much dislike the use of the 'this' pointer, unless passing this object to another method. Since pipScale and the like are private final fields, they should be declared as mPipScale, etc which will eliminate this need.

rajawali/src/main/java/org/rajawali3d/renderer/RajawaliPipRenderer.java
```diff
@@ -0,0 +1,192 @@
...
+        this.pipMarginX = pipMarginX;
+        this.pipMarginY = pipMarginY;
+    }
+
+    public void setMiniRenderer(SubRenderer mMiniRenderer) {
+        this.mMiniRenderer = mMiniRenderer;
```
jwoolston: Conversely to the above, the parameter `mMiniRenderer` should not have the 'm' prefix as it indicates a class internal field. `this.mMiniRenderer` is fine, but the `this` can be removed once the parameter name is changed.

rajawali/src/main/java/org/rajawali3d/renderer/RajawaliPipRenderer.java
```diff
@@ -0,0 +1,192 @@
...
+        mMainRenderer.initScene();
+
+        // Init mini scene
+        mMiniRenderer.initScene();
+    }
+
+    @Override
+    protected void onRender(final long ellapsedTime, final double deltaTime) {
+        try {
+            // Render mini scene into mini quad
```
jwoolston: Thankyou for adding comments to indicate the process.

jwoolston:
Added a few requests for some naming convention stuff. Otherwise looks good. Thanks for documenting it. As for why it built, its because you kept your change set confined to the change you actually made rather than including all kinds of other edits and formatting changes to the existing code base.

adamantivm:
@jwoolston @ToxicBakery thank you very much for your kind review. I agree with and address your comments. I also fixed a small bug in the calculation of the mini scene.

rajawali/build.gradle
```diff
@@ -29,11 +29,19 @@ android {
             ext."signing.password" = System.getenv('SIGNING_KEY_PASSWORD') != null ? System.getenv('SIGNING_KEY_PASSWORD') : System.properties['SIGNING_KEY_PASSWORD'] ? System.properties['SIGNING_KEY_PASSWORD'] : hasProperty('SIGNING_KEY_PASSWORD') ? property('SIGNING_KEY_PASSWORD') : ""
             ext."signing.secretKeyRingFile" = "./rajawali_secret.gpg"
 
-            version = rootProject.getReleaseVersion(property('VERSION_NAME'))
+            // version = rootProject.getReleaseVersion(property('VERSION_NAME'))
```
adamantivm: argh, this is an accidental add. I'll remove this in a minute.

PR #1548, opened 2015-09-04
update

ravenfeld:


ToxicBakery:
I'm not really sure what this PR is doing. If this is something you would like merged can you please make a new branch to PR against with clean commit history? Ideally just a single commit. Also please avoid including things like the IML files. We need to remove and ignore those if any are still in the repo.

PR #1553, opened 2015-09-17
Update Gradle plugin and compile SDK version

vkotovv:
Update target and compile sdk to 23.
Fix Travis config.
Had to add legacy alias because of this : http://stackoverflow.com/questions/31433687/android-gradle-apache-httpclient-does-not-exist


ToxicBakery:
Thank you!

PR #1583, opened 2016-02-24
Fix/RajawaliRenderer:unProject()

yolandacorreia:
. Creating a copy of the projection matrix, so it's not altered when multiplied by the view matrix on ``RajawaliRenderer:unProject()`` function call


jwoolston:
The build failure is not due to anything wrong with the PR.

Issue #1504, opened 2015-06-08
"createContext  failed: EGL_BAD_ATTRIBUTE" exception on Samsung S3


rpicolet:
master Snapshot 267

When the GL thread starts up, it throws an exception (stack trace below).  It appears to be caused by Capabilities.getGLESMajorVersion() returning version 3 (instead of version 2), as I can make the 
problem go away by replacing RajawaliSurfaceView with a version that hard-codes 		setEGLContextClientVersion(2);

Don't know if it's related, but prior to the GLThread start [after setSurfaceRenderer() during onCreateView()], a warning appears (also pasted below) from the Adreno driver about  "not forcing ES30"...

== Exception stack ==

06-07 18:30:51.200  19489-19524/shgames.stu W/Adreno-EGL﹕ <qeglDrvAPI_eglCreateContext:2407>: EGL_BAD_ATTRIBUTE
06-07 18:30:51.200  19489-19524/shgames.stu W/dalvikvm﹕ threadid=19: thread exiting with uncaught exception (group=0x41734da0)
06-07 18:30:51.210  19489-19524/shgames.stu E/AndroidRuntime﹕ FATAL EXCEPTION: GLThread 14532
    Process: shgames.stu, PID: 19489
    java.lang.RuntimeException: createContext failed: EGL_BAD_ATTRIBUTE
            at android.opengl.GLSurfaceView$EglHelper.throwEglException(GLSurfaceView.java:1202)
            at android.opengl.GLSurfaceView$EglHelper.throwEglException(GLSurfaceView.java:1193)
            at android.opengl.GLSurfaceView$EglHelper.start(GLSurfaceView.java:1043)
            at android.opengl.GLSurfaceView$GLThread.guardedRun(GLSurfaceView.java:1410)
            at android.opengl.GLSurfaceView$GLThread.run(GLSurfaceView.java:1249)

==== Adreno Warning ====

06-07 18:30:50.319  19489-19489/shgames.stu W/Adreno-GSL﹕ <get_panel_settings:3802>: Android framework reported version 2. So, don't force ES30
06-07 18:30:50.329  19489-19489/shgames.stu I/Adreno-EGL﹕ <qeglDrvAPI_eglInitialize:381>: EGL 1.4 QUALCOMM build: AU_LINUX_ANDROID_KK_2.7_RB1.04.04.00.007.018_msm8960_KK_2.7_RB1_CL3869936_release_AU (CL3869936)
[...]

Issue #1387, opened 2015-03-17
Children of Object3D are not pickable with ObjectColorPicker


C0br4:
When a child of an `Object3D` is registered in an `ObjectColorPicker`, it can't be picked. 
As aworkaround I remove the child from its parent `Object3D` and add it to the `RajawaliScene` directly which weirdly works.

jwoolston:
This is probably because child objects arent exactly treated the same from the standpoint of uniqueness. I would assume off hand that touching child or root objects would result in the root being chosen - not sure that that is what happens either though.

C0br4:
@jwoolston The `onObjectPicked()` from the corresponding Listener is never called.

jwoolston:
@MasDennis do you have any input on this?

MasDennis:
I'll investigate

rpicolet:
I forked a repo in order to fix this (and a couple of other minor problems).  I got things working for my use cases, but the scope of the changes - while not huge - was not trivial either, and right in the heart of all the rendering logic (modified Material.java, Object3D.java, ObjectColorPicker,java, and RajawaliScene,java).  I am quite willing to submit a pull request if someone on the team has time to review/test my changes against other use cases, but I don't want to claim I understand the risks...

jwoolston:
@rpicolet If you have a branch with a clean set of changes to fix this I can take a look, just submit a PR and I should be able to pull the code locally.

rpicolet:
PR coming soon. I will document my concerns there.  Thanks!

jwoolston:
Should be resolved as of #1614

Issue #1580, opened 2016-02-24
Desmurf the class names


jwoolston:
Over the years, Rajawali has suffered from Smurfing Syndrome. I am at least half at fault for this so I will correct it. 

This will modify the public API and the version number will be changed accordingly.

(Smurfed class names are when they all start with the same word, in this case, Rajawali).

PR #1586, opened 2016-02-25
#1387 pull request 

rpicolet:
# High-level Change Summary

## RajawaliScene
* Abstracted/separated object-picking render pass into `doColorPicking()`, attempting to isolate it from the bulk of `render()` control flow for primary (visible) render passes and plugins  
* As picking request is based on prior frame state, called `doColorPicking()` first thing in `render()` before any frame updates
* Factored out some code common to both `render()` and `doColorPicking()` into `configureTargetRendering()`
* Fixed counter on preDraw callback processing, normalized naming for FrameCallbacks and counters 

## Object3D
* Defined/applied a default `UNPICKABLE` picking color to all Object3Ds
* Added `isColorPicking` parameter to `render()`, which always processes entire Object3D tree (similar to 0.9)
* All visible Object3Ds get rendered during color picking, not just those registered for picking, to allow for obscuring but unpickable foreground objects

## ObjectColorPicker
* Changed `unregisterObject()` to reset color to UNPICKABLE
* Deprecated/renamed `createColorPickingTexture()` to `pickObject()`

# Concerns

## Logic & Performance
* Guessing that `Object3D.render()` call of `preRender()` should be skipped for color picking pass? 
* Not sure if normal (non-color-picking) Skybox rendering should use `sceneMaterial` (was set to null before, don't remember why I changed it)
* Not sure if textures are properly bypassed/replaced during color-picking render
* Not sure about performance cost for rendering all objects during color picking

## Footprint, Testing, & Merging
* When I started out on this, I wasn't at all sure I would be able to figure it out, and wasn't really thinking about merging and white-space/name changes/miscellaneous minor code cleanups. so there are a number of these that don't really heed to be there; if not acceptable, I'll try and back most of them out
* Beyond those, placing `doColorPicking()` before any frame updates and rendering all visible objects during color picking are not really within the scope of #1387, but I think they are in the spirit of providing a reasonable color-picking function overall...
* I also didn't post an issue for the preDraw counter bug...
* I am not confident of the approach I used to test my changes against RajawaliExamples:
  * Downloaded/expanded RajawaliExamples.zip
  * When I tried to run it from AS on my Samsung S3, got the same error I reported in issue #1504, so I modified my local versions of `RajawaliSurfaceView.initialize()` and `RajawaliTextureView.initialize()` to use a hard-coded version 2:

`//	final int glesMajorVersion = Capabilities.getGLESMajorVersion();` 
`        final int glesMajorVersion = 2;`

  * Added an .aar import module to the Examples AS project using the resulting rajawali-debug.aar, and replaced the dependency in examples.build.gradle 

* I then was able to run most of the examples on the S3, but the following examples apparently fail, and I am not sure if any of these are due to my changes or approach to testing or not (although I'm very suspicious of all the filter/effect/post-processing failures...):
  * GCode Toolpaths
  * Coalesce Animation
  * Animated GIF Texture
  * Sepia Filter
  * Greyscale Filter
  * Gaussian Blur Filter
  * Multi Pass
  * Bloom Effect 

* If I have missed the boat on any significant issues, please feel free to ask me anything, send it back with change requests, make any improvements you think appropriate, or deny the pull request altogether; I'll survive :-)

Thanks, and make sure to have a beer before you take any of this on!


jwoolston:
> Thanks, and make sure to have a beer before you take any of this on!

Yea this is a serious PR. Please don't be offended if it takes me a few days to get through this. I will also want some input from the other guys before a final merge. Solid, solid write up though, thankyou. 

I realize it happened in the middle of you putting this together, but #1580 changed the class names in `master` to remove the Rajawali prefix. I don't mind sorting out the conflicts if needed but if you feel comfortable pulling master into this branch, resolving the conflicts and pushing another commit to this pr so it doesn't have conflicts, that would be helpful. Again, since I modified the API, I don't mind fixing the conflicts.

rpicolet:
Merging my fork's master branch with Rajawali master is complete, and wasn't hard.  The only file that required manual merging was RajawaliScene (which makes me wonder if it was intentionally non-de-smurfed for some reason...)

Is the "All checks have failed" message due to something I did wrong?  Let me know if so and I will try again.

I understand this PR could take a while for the team to analyze, and that it may not even pass muster.  No hurry/pressure from my end, my app is proceeding nicely, just thought I'd see if my solution would be general/robust enough to address #1387 and possibly save some duplicate effort.

jwoolston:
> which makes me wonder if it was intentionally non-de-smurfed for some reason

Nope, just missed. Its corrected on `development` already.

> Is the "All checks have failed" message due to something I did wrong? Let me know if so and I will try again.

No it is not. I am not certain why but there is some issue with finding the signing key entries in the travis config, but only for PRs. I suspect it has something to do with the old travis.yml files people have in their forks, but I would have thought the merges would take care of that.

rpicolet:
OK, thanks.

rajawali/src/main/java/org/rajawali3d/Object3D.java
```diff
@@ -48,6 +48,8 @@
     public static final int BLUE = 2;
     public static final int ALPHA = 3;
 
+	public static final int UNPICKABLE = Color.argb(255, 255, 255, 255);
```
jwoolston: Since color picking doesn't have much meaning with transparent objects, perhaps it would be better to use a 0 alpha channel, allowing us to use white as a picking color.

rpicolet:
Picking-color values are `int`s auto-incremented from 0 so that they can be used as direct lookup indices after reconstruction from the rendered pixel data.I didn't change any of that, and didn't think about a use case where picking-colors would be specified instead of generated as indices. 

I coded `UNPICKABLE = Color.argb(255, 255, 255, 255)` instead of `UNPICKABLE = -1`; either way the value could never be used as an index. It's also the same as the background color used for the color-picking render pass...

jwoolston:
OK, works for me. I've never been 100% familiar with how the color picking part of the library worked.

rpicolet:
Picking-color values are `int`s auto-incremented from 0 so that they can be used as direct lookup indices after reconstruction from the rendered pixel data. I didn't change any of that, and didn't think about a use case where picking-colors would be specified instead of generated as indices. 

I coded `UNPICKABLE = Color.argb(255, 255, 255, 255)` instead of `UNPICKABLE = -1`; either way the value could never be used as an index. It's also the same as the background color used for the color-picking render pass...

rpicolet:
It was a first for me too...  Thanks for looking at it and challenging my assumptions.

rajawali/src/main/java/org/rajawali3d/Object3D.java
```diff
@@ -174,30 +175,50 @@ protected void preRender() {
...
-			final Matrix4 parentMatrix, Material sceneMaterial) {
+			final Matrix4 parentMatrix, Material sceneMaterial, boolean isColorPicking) {
+
 		if (!mIsVisible && !mRenderChildrenAsBatch)
 			return;
 
 		Material material = sceneMaterial == null ? mMaterial : sceneMaterial;
-		preRender();
+
+        if (!isColorPicking) {
```
jwoolston: I would like to see `preRender` always be called. In its current form, it validates that the VBOs have been created for the object. This could potentially be necessary for any object which is added after scene initialization but which was created without initializing its buffers (necessary to create objects on a thread other than the GL thread).

jwoolston:
> Not sure about performance cost for rendering all objects during color picking

Regarding this issue, the performance cost is of course variable, however, you state the following in your justification for it:

> All visible Object3Ds get rendered during color picking, not just those registered for picking, to allow for obscuring but unpickable foreground objects

This is a very valid use case to consider, so I think the impact is worth it. The reality is color picking is not the most efficient way to do object picking so this is a justifiable trade off.

rajawali/src/main/java/org/rajawali3d/util/ObjectColorPicker.java
```diff
@@ -71,24 +71,30 @@ public void unregisterObject(Object3D object) {
...
 	public void getObjectAt(float x, float y) {
-		mRenderer.getCurrentScene().requestColorPickingTexture(new ColorPickerInfo(x, y, this));
+		mRenderer.getCurrentScene().requestObjectPicking(new ColorPickerInfo(x, y, this));
 	}
 
 	public RenderTarget getRenderTarget() {
 		return mRenderTarget;
 	}
 
+	@Deprecated
```
jwoolston: Ditch this method completely. I just made a fairly substantial change to the API and this is a trivial adjustment to make.

jwoolston:
@rpicolet did you notice any odd behavior with the lights? In particular, when I test the merged version (Artifact `org.rajawali3d:rajawali:1.1.346-SNAPSHOT@aar`) the Object Picking example with monkey heads has no lighting applied other than the ambient.

jwoolston:
There seem to be a fair number of issues, and I doubt all, if any, are related to the changes here. Color picking itself seems to work fine and in the interests of seeing this contribution applied, I will merge once the issues I have pointed out are corrected.

rajawali/src/main/java/org/rajawali3d/Object3D.java
rpicolet: I wasn't (and am still not) sure about the intended uses for `preRender()`; I didn't (at the time) find any usages in the library, so it seemed like a client hook, so I was concerned about having some way of indicating the rendering pass/context under which it was being called (object picking vs visual render passes) in case it involved a lot of processing or memory, but didn't want to modify its signature to pass along some kind of indicator.

Validating VBOs certainly sounds worthwhile, but maybe a task that does that could be queued to the GL thread as part of addChild() so it doesn't have to be checked on every render pass? just brainstorming. Any which way, I defer completely to your opinion here.

rajawali/src/main/java/org/rajawali3d/util/ObjectColorPicker.java
rpicolet: Sure, no problem.

jwoolston:
All of the post processing effects other than Shadow Mapping (it was already broken) seem to break with this. I suspect it has something to do with the changes of passing scene material around.

rpicolet:
I've been wandering around in the postprocessing stuff to try and understand how those failures might be related to my changes.  In doing so, I found a couple things that seem problematic or revealing:

* Some of the Pass-flavored class render() methods make gl calls - e.g. glClearColor() - and assume those will not be changed by scene.render(), but the object picking pass could write over them. This might explain some of the current failures, but it seems like it should only cause problems when object picking has been requested, which it clearly isn't in the failing Examples; still it looks like an integration headache yet to be solved (maybe with centralized render pass management under the Renderer?)

* If I'm reading things right, it seems like PostProcessingManager,updatePassesList() might skip some passes, as APass.getType() returns PASS and APostProcessingEffect.getType() returns MULTIPASS, but updatePassesList() only checks for PASS and EFFECT, and I couldn't find anything that returns EFFECT...  Can't really see how this might be affected by object picking, but seems like it might explain some of the failures...

BTW, in addition to the bad ligthing you mentioned and which I also observed, the ObjectPicking example is also supposed to rotate the monkeys, and that isn't happening either. Not sure if that is a clue or unrelated; hard to figure how I could have broken that (animations work just fine in my app).

All of which begs the question of whether it would be simpler to just abandon my color picking solution altogether and focus on a different one...  Still OK by me if that makes more sense to the team. In the meantime, I'll try to continue investigating.

jwoolston:
Regarding the animations, I'm certain that is not related to your changes. Some of the lighting issues I saw as well I believe based on some work yesterday were already present so dont worry about those. The post processing is my primary concern. I will try and step through the code tomorrow and take a look. One thing to keep in mind with the post processing is it is all intended to be a separate render pass, as is the color picking, so that could be where the issue is coming from. As for the git stuff related to the other issue, I will try and walk you through a solution tomorrow. I was thinking about cherry picking that one commit but if I do it will complicate this PR substantially.

rpicolet:
Sounds good, I need to improve my git skills anyway, so let me figure it out and I'll just do it right; can't be that hard. You have bigger fish to fry, so don't waste any more time holding my hand on that stuff.

jwoolston:
@rpicolet Sure. To help give you a head start what we want is for there to be 3 branches on your machine:

- A clean branch tracking Rajawali/Rajawali/master either through your fork or directly
- A branch with the color picking changes (new pr from this)
- A branch with the GL ES version handling (second new pr from this)

We will close this pr then. The best way to accomplish this will be to create a new branch from your change set, say `changes-preserve`, then reset your `master` to the Rajawali `master` `HEAD`. From here you create your working branches, and then cherry pick the necessary commits over. Conflicts may arise during this process that you will have to sort out but it shouldn't be too crazy, just tedious. 

To prevent this in the future, never develop on the `master` branch. As a general rule of thumb, if you are are adding, fixing, tweaking, removing or in general doing anything to the code, create a branch. 

Get through this, and you will have passed the GIT right of passage. It seems you are committed to seeing this through (which I think is fantastic) and you are capable so a premptive welcome to the club is in order! :beers:

rpicolet:
Excellent guidance, that maps out where I want to go, so I can now just figure out how to git there...

I'll probably be heads down for a couple days with this, so you will have a break from my meanderings for a while.

And thanks for the beers!

jwoolston:
No problem. Feel free to ask if anything in git rears its ugly head. Some of the more low level branch manipulation stuff can get pretty nasty. Just make sure you copy your branch as is and dont push anything until you think you have it figured out, that way you wont loose anything.

rpicolet:
Yep!

rpicolet:
Per above discussions, about to close this PR. 

Since there are only 3 files changed for this PR and one for the #1504 fix, and I have not been using branches for any of them, my plan is to simply make local (& backed-up) copies of my 4 changed files, delete my fork and local repos, start over with a fresh fork and clone of the latest Rajawali master, and proceed with the normal forked repo/branched development workflow, as Jared described above.  For a collaborative Git rookie like me, I think this will give a cleaner result than trying to repair/revert/reset/rebase/whatever my current repos.  I don't think the loss of history in my repos will matter to anyone?

Just waiting for a go/no-go...

rpicolet:
Oh yeah, I don't have to actually delete the old repos.  Just start with some fresh ones, wait to delete the old ones until everything is tested/pulled/merged back to Rajawali/Rajawali...

rpicolet:
So I'm going ahead with it...

jwoolston:
The history will really only matter to you. In general, we prefer change sets to be confined to a single commit anyway, though we are not hard core about that, so go with whatever process you are comfortable with that gets the changes separated out.

rpicolet:
Well, after studying git workflows and drinking the branch-only-dev koolaid, I went ahead and started with a fresh github fork, and a local clone repo (that tracks Rajawali/master). I was able to successfully build the `rajawali` module, and I would like to make sure I can run the Examples app before I start making any of my branches/changes, but the `vuforia` module (and so the `examples` module) failed due to a failed `ndk-build`:

>...
>:vuforia:ndkBuild FAILED

>FAILURE: Build failed with an exception.

>* What went wrong:
>Execution failed for task ':vuforia:ndkBuild'.
> A problem occurred starting process 'command 'ndk-build''

>* Try:
>Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

>BUILD FAILED

>Total time: 0.628 secs 

I do have the NDK installed, with my `PATH` updated accordingly, and it shows up OK in the AS terminal in response to a `which ndk-build`, but past that I am stuck for how to proceed, as I am unsure where/how to insert any of the suggested run options (--stacktrace etc).  

I did notice that `vuforia/build.gradle` has a comment that seems like it might be related:

`jni.srcDirs = [] //disable automatic ndk-build call`

Also, `RajawaliVuforia.cpp` has major editor complaints about all of its JNI calls, as if something is not configured correctly...

Any suggestions much appreciated!

jwoolston:
The editor complains are due to marginal support in AS for NDK development and can safely be ignored. 

As for the ndk build, I suspect you have a slightly stale version of the branch. Late last night is when I finally got the Travis build for it fixed and it was failing because I had the gitignore file configured incorrectly and there were source files missing. Please confirm you have the latest and I apologize for causing the confusion.

rpicolet:
master HEAD = d3f9496...

jwoolston:
hmmmm. I'll have to look into it more. In the meantime, open settings.gradle in the root and comment everything other than examples and rajawali. Then, just dont commit that file.

jwoolston:
@rpicolet Do you have an ndk.dir entry in your local.properties file?

rpicolet:
No, only an `sdk.dir`. Taking the hint, I tried adding an `ndk.dir` entry that points to my NDK install (even though its a generated file), but unfortunately it made no difference...

I also now see a lot of lint warnings (didn't know the underline indicated anything 'til I accidentally cursored over some of them) in `vuforia/build.gradle`; for example, `VERSION_NAME`, `buildNumber`, and `jniLibs` all indicate "Cannot resolve symbol", and `ndkBuild` and `commandLine` both indicate "Cannot infer argument types"...

BTW, `examples/build.gradle` has `compile project(':vuforia')` in its dependencies, so it doesn't build without vuforia, so changing settings.gradle doesn't work...  I can build the rajawali module/library directly, just not examples, and I just wanted to be sure I have a clean test bed before starting my branches...

rpicolet:
Oh, I see the lint warnings in rajawali/build.gradle too, and it works fine, so that's probably not the issue...

jwoolston:
AS doesnt have great gradle support yet either so even though gradle can see those values from the gradle.properties files, AS thinks they dont exist. I have the ndk installed via the ndk bundle option in the SDK managers 'extras' section. After doing that, my local.properties file looks like:

````gradle
ndk.dir=C\:\\Users\\jared\\AppData\\Local\\Android\\sdk\\ndk-bundle
sdk.dir=C\:\\Users\\jared\\AppData\\Local\\Android\\sdk
````

On the command line or via a custom gradle command in AS can you try running

````
gradlew :vuforia:ndkBuild --stacktrace
````
And let me know what it says?

rpicolet:
Ah, I wasn't aware of the new ndk bundle approach.  FYI, and for reference by other collaborators, on my Mac, the install option shows up only under Tools->Android->SDK Manager->SDK Tools as Android NDK, and not at all with the old standalone SDK Manager (which is why I originaly went looking for it elsewhere). In any case, I un-installed my prior external NDK, and installed the new bundle accordingly and now see a corresponding entry in local.properties:

>ndk.dir=/DevApps/android-sdk-macosx/ndk-bundle
>sdk.dir=/DevApps/android-sdk-macosx

and under project structure I still see Gradle version 2.10 and Android Plugin Version 1.5.0

Unfortunately, that made no difference; it still fails the same way.  Running from the AS terminal window command line per your suggestion also made no difference (same failure point and output as above), and apparently yielded no stack trace (likewise, I saw no additonal output running with the --info or --debug options... )

Been looking around on stackoverflow and Android Tools Project SIte, etc, I'll continue to scratch my head along with you.  Let me know if you think of other experiments/ideas, and I will do likewise.

BTW, I am blown away by all the improvements being made to Rajawali 1.1; it will be an awesome release!

jwoolston:
What happens if you run this on the terminal:

`ndk-build V=1 -C path-to-vuforia/src/main`

jwoolston:
@rpicolet to help speed things, feel free to find me on Google Hangouts. The email in my profile will connect.

rpicolet:
Been running experiments, discovered several things:

* Using full paths on the command line for both `ndk-build` and `vuforia/src/main` worked, which proves the basic make is working, and the issue I'm having is somewhere in my AS/Gradle configuration.

* My earlier attempts to run `gradlew` with a stacktrace were invalid. Per the Gradle Wrapper docs, i should have been using `./gradlew`; when I tested that with `./gradlew -v`, it first proceeded to download/unzip/install the wrapper distribution, before displaying the version, and so when I ran `./gradlew :vuforia:ndkBuild --stacktrace` it actually produced a trace:

       [gradlew stacktrace.txt](https://github.com/Rajawali/Rajawali/files/162541/gradlew.stacktrace.txt)

* As I'm sure you can see, the stack trace shows that the AS/Gradle plug-in build can't find the `ndk-build` command; I proved this is the problem by adding the same path as in the local.properties `ndk.dir` to my environment's `PATH`, and then basically building the whole project from a simple `./gradlew examples:assemble` command.

* Unfortunately, I still can't download/run the examples app from AS. I haven't looked into running it from the command line yet, procrastinating/wondering if there is a way to reference the `ndk.dir` value directly in the `ndkBuild` task or some other straigh-forward fix for the whole problem?

Thanks for the Hangouts suggestion, I'll sign up :smile:

rpicolet:
Here's the build output...

[gradlew build.txt](https://github.com/Rajawali/Rajawali/files/162605/gradlew.build.txt)

rpicolet:
Sorry to disappear for the last few days. 

I reapplied my changes for this feature to a clean branch, and as before, object picking works.  But I now also know my changes break many post-processing examples.  I spent quite a while stepping through the `GreyScaleFilterFragment` example to see if I could isolate the cause(s?), and while I still have one lead to follow-up (`ScreenQuad` sets both `mEnableDepthTest` and `mEnableDepthMask` to false, which might interact with my code in unexpected ways, still looking at it), i have become increasingly worried about whether the feature/branch should be merged.

My implementation is basically a refactoring of the 0.9 implementation of color picking, written in ignorance of and completely at odds with the multi-pass rendering approach of the `PostProcessingManager` (and I assume, vice versa).  While I attempted to isolate the color-picking render pass from the rest of the rendering passes as best I could, as the broken post-processing shows, there are still some unintended interactions. And after looking at many of the post-processing use cases, I am not at all convinced that my code won't continue to unduly complicate integration for the general Rajawali library going forward, especially as other post-processing elements mature; it's all just very tightly coupled with multiple competing control flows through the `Scene` and `Object3D` `render()`s.

Assuming color picking itself is still a desirable long-term goal (as opposed to implementing `RayPicker`?), I am wondering if a different approach altogether might be cleaner.  Perhaps as yet another `IPass` with `PassType.COLOR_PICKING`, and with an `ObjectColorPickingPass` extending `APass`.  Of course, such an approach would raise other issues, such as how the pass would get set up and what if any changes to the public API would be desirable.  It would probably also creep into broader library questions like whether merging the `PostProcessingManager` directly into the `Renderer` might have some benefit, whether frame-level functions/callbacks should be separated out from the render passes, whether some internal-only `render()` hooks/wrappers might yield some wins, and so on.  

That all seems pretty ambitious/ risky, so even assuming such an approach would appeal to the team and someone has the time to work on it (I'd be willing to put together an informal spec to bootstrap some discussion and see where things go), it seems like it would probably be well out of scope for Bombshell.  But I thought I would throw my musings out and see what comes back.

In the meantime, if it still seems useful to proceed with my current implementation (even if it gets backed out later), let me know.  I can submit a tentative PR (that breaks the post-processing, unless I find/fix the problem soon) for you and the others to consider.

Notice how I well I am avoiding future blame?  :grin:

jwoolston:
@rpicolet Your efforts here are greatly,greatly appreciated. Due to the nature of how color picking works, I am in favor of the additional post processing type. I would like to see color picking working at a minimum for Bombshell. There are reasons why it should be preferred over ray picking...they each serve their purpose and they're purposes each differ. If you could create a ticket detailing what you have in mind, focusing on the minimum functionality required to get color picking functioning, we can focus the discussion there.

rpicolet:
OK, will do.  By ticket I assume you mean a new issue?  Let me scratch my head for a day or two, and I'll post a draft to get things started.

I'll start with the assumption that the current `ObjectColorPicker` usage is still the desired client interface, and that integrating the `PostProcessingManager` pass-management functions into the `Renderer` is OK, even though that will change the post-processing usage model (i.e. no more creating a `PostProcessingManager` or overriding `onRender()`).  If that's not the general direction you are thinking of, just let me know so I don't waste anyone's time.

jwoolston:
> I'll start with the assumption that the current ObjectColorPicker usage is still the desired client interface

Yes, this is a good place to start. We can edit the API from there.



> i.e. no more creating a PostProcessingManager or overriding onRender()

Lets not do that. At least initially, I see no reason why we cant go down the road of keeping it the same post processing model. If a problem arises in your initial framing, stop, post what you have and the issue, and lets discuss from there. The current post processing system was a lot of work and should be pretty scalable and adaptable, so lets not jump to altering it drastically just yet. If you feel there is a strong reason to do so, again post it and we can discuss. @ToxicBakery and @MasDennis will surely have useful contributions to the discussion from different viewpoints than myself or you, as will other Rajawali users should they decide to comment on the discussion. Here in lies the strength weakness beauty and ugly of open source.

rpicolet:
Thanks, I look forward to everyone's input, as I'm positive I do not understand all the issues, and have much to learn. I am _not_ a 3D graphics/GL expert in the least, that's why I chose Rajawali for my app development several years ago.  I also understand that I am mostly an unknown quantity to everyone on the team, so please don't worry about challenging my suggestions or assumptions, make me earn your respect, and by all means suggest better ideas. I never come up with my own best idea in the first few tries, and that's when I'm in familiar territory. My above proposal was a premature solution brainstorm, but at least it got things rolling!

It's probably better to start by nailing down the basic goals/requirements. I propose that these are:
* to allow all 4 combinations of the two features to inter-operate reliably: default rendering, default rendering plus color picking, post-processing, and post-processing plus color-picking.
* to better isolate the very different "never visible" color-picking GL render details from those of the "visible" renders, and so make it easier to test/maintain both, and to add new features to either
* to minimize changes to the client API for both features,
* within reason, to maximize code re-use and minimize the code footprint

I'll wait to get some feedback on these goals before doing anything more on this. I should say, even if most everyone agrees with these as the goals, I haven't really yet thought of an approach that accomplishes all of them well, so getting some sense of relative priorities would also help me a lot in proposing a design.

BTW, should we start a new issue, or is this still the right place?

jwoolston:
> BTW, should we start a new issue, or is this still the right place?

This should all be in a new issue. Please copy your proposal over to a new issue for commentary.

PR #1587, opened 2016-02-25
Added segmentation to screen quads.

contriteobserver:
So we can use custom vertex shaders effects.

This is my first pull request, please let me know if I have not followed any of the normal process.


jwoolston:
The build failure has nothing to do with the PR.

jwoolston:
I have some reservations about this as the purpose of a screen quad is specifically to render a texture to the screen....that said, the changes should still allow for that use while adding additional functionality. I will test and validate. Id also like @MasDennis to weigh in.

MasDennis:
@jwoolston @contriteobserver Sorry for the late reply! I'm fine with this change, I can see where it could be useful.

PR #1600, opened 2016-03-08
PR for #1504

rpicolet:
See commit message...

If the secondary check (for color bit sizes) is not useful as-is, I'll be happy to remove/modify it and retest.
I ran all the examples on my S3 and noted anything unusual in the attached file:

[issue-1504 example notes.txt](https://github.com/Rajawali/Rajawali/files/163926/issue-1504.example.notes.txt)

Of particular interest: ETC2 provides an appropriate message that ES 3 is not supported, and a Vuforia crash.

As an aside, libVuforia.so is getting (re)built every time even though it is under source control, and so shows up as modified. Is this intentional?

Let me know if there are any git/branching issues I've missed.

### DIFF
--- a/rajawali/src/main/java/org/rajawali3d/util/Capabilities.java
+++ b/rajawali/src/main/java/org/rajawali3d/util/Capabilities.java
@@ -72,28 +72,49 @@ public class Capabilities {
         final EGLDisplay display = egl.eglGetDisplay(EGL10.EGL_DEFAULT_DISPLAY);
 
         final int[] version = new int[2];
-        if (!egl.eglInitialize(display, version)) throw new IllegalStateException("Failed to initialize and EGL context while getting device capabilities.");
+        if (!egl.eglInitialize(display, version))
+            throw new IllegalStateException(
+                    "Failed to initialize an EGL context while getting device capabilities.");
         mEGLMajorVersion = version[0];
         mEGLMinorVersion = version[1];
         // RajLog.d("Device EGL Version: " + version[0] + "." + version[1]);
 
-        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.JELLY_BEAN_MR2) {
-            // The API for GLES3 doesnt exist on this device
-            // RajLog.d("Device is API level cannot support OpenGL ES 3.");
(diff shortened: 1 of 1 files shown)


jwoolston:
> If the secondary check (for color bit sizes) is not useful as-is, I'll be happy to remove/modify it and retest.

No, it is not commonly needed, but we should leave it.

jwoolston:
> Of particular interest: ETC2 provides an appropriate message that ES 3 is not supported, and a Vuforia crash.

The vuforia examples are unvetted right now so don't count them as a problem. In particular, they will always fail on Marshmallow devices. 



> As an aside, libVuforia.so is getting (re)built every time even though it is under source control, and so shows up as modified. Is this intentional?

`libVuforia.so` is a closed source library provided by Qualcomm. It is not being rebuilt each time but being linked into `libRajawaliVuforia.so` which is rebuilt each time and is not under source control.

rpicolet:
>libVuforia.so is a closed source library provided by Qualcomm. It is not being rebuilt each time but >being linked into libRajawaliVuforia.so which is rebuilt each time and is not under source control.

It's showing up as modified after I do a build: 

>Randys-MacBook-Pro:Rajawali Randy$ git status
>On branch issue-1504
>Changes not staged for commit:
>  (use "git add <file>..." to update what will be committed)
>  (use "git checkout -- <file>..." to discard changes in working directory)
>
>    modified:   vuforia/src/main/libs/armeabi-v7a/libVuforia.so
>
>no changes added to commit (use "git add" and/or "git commit -a")

Not sure why linking it into libRajawaliVuforia.so would change it? Oh! just noticed there are two copies, one under src/main/jni and one under src/main/libs; I bet ndk-build copies the (original?) one in jni over to libs, so that copy probably doesn't need to be under source control... ??

jwoolston:
Yes. It should be removed from tracking and the gitignore updated for that one only.

rpicolet:
Looks like it's already in .gitignore...

rpicolet:
No, I'm wrong, it's not there yet, just the RajawaliVuforia lib.

Issue #1595, opened 2016-03-03
Migrate Rajawali modules to main repository


jwoolston:
All of the various Rajawali modules in other repositories should be migrated to the main repository as AS modules. This allows us to deploy artifacts for all with any updates, and ensures that any changes to the core Rajawali library are affected across all artifacts while also facilitating simpler testing.

- [x] Examples App
- [x] VR 
- [x] Vuforia
- [x] Wear
- [x] Wallpaper (Migrated, needs debugging)
- [x] Update all the repository README.md files.

PR #1602, opened 2016-03-09
Adds wear support

jwoolston:
#1595 Adds a wear library and example project.


ToxicBakery:
I think you need to add `- extra-google-m2repository` in the components section of travis to get the wear dependencies.

jwoolston:
> I think you need to add - extra-google-m2repository in the components section of travis to get the wear dependencies.

https://github.com/Rajawali/Rajawali/blob/master/.travis.yml#L15

Already there.

ToxicBakery:
No that is the android m2 not the google m2

jwoolston:
:sigh: why cant they just put the damn things in jcenter.

ToxicBakery:
You and everyone else wants to know why too. It's really bloody annoying the update tool is garbage and you are forced to use it.

Issue #1550, opened 2015-09-11
Object3D.setScreenCoordinates() anomaly?


rpicolet:
For this method to correctly display an Object3D at screen coordinates (x, y) it seems to be necessary to specify the y complement of the screen height, in other words (x, viewportHeight - y), undoing the same expression in the method's implementation...

Not sure if this is intentional, and easy enough to workaround, just thought I'd point it out in case the API needs to change...

org.rajawali3d:rajawali:1.0.279-SNAPSHOT@aar

jwoolston:
I believe it is unintentional as I encountered some odd behavior with that function recently. Ill bump this to a bug to look into it at some point. Thanks.

rpicolet:
This should obviously be low priority.

Thanks for all your hard work and continued support, I know many of us users can seem pretty demanding without contributing all that much.

jwoolston:
> Thanks for all your hard work and continued support, I know many of us users can seem pretty demanding without contributing all that much.

As long as the requests/bug reports that come in are politely worded, it doesn't bother me. Obviously there is only so much that can be done in our limited time. The ones that annoy me are when no attempt has been made to ensure that there is actually a bug or when they are worded to imply we must fix it immediately because of some timeframe applying to the OP. In this case, the workaround you present indicates it clearly is not functioning as intended.

jwoolston:
I took a closer look at this today an I believe it is functioning as intended. The reason being, Open GL uses a different coordinate system than Android screen points. In OpenGL, +Y is at the top of the screen, while in Android it is at the bottom. Were you taking this into account?

rpicolet:
No, I was confused and just asking the question.  A javadoc update would work fine to clarify.  As with a number of methods, it's not always clear to a newbie like me which coordinate system/units/constraints are assumed.  So in your copious free time... 

Just kidding.  Not sure what the plans are for docs, but I would consider pitching in for a beer or two...

jwoolston:
The plans for docs are, well, wishful thinking. Development tends to take precedence. Some of the devs and I meet up in SF from time to time for beers. I'll have to get you on the next email chain for it.

rpicolet:
Sounds good, I'm gettin' a little thirsty...

ToxicBakery:
We gladly accept pull requests for documentation ;)

PR #1604, opened 2016-03-09
PR Fixing #1550 - Object3D.setScreenCoordinates() anomaly?

rpicolet:
Documentation only

Any feedback is welcome: accuracy, vocabulary, style, completeness, verbosity; praise, condemnation, laughter...

### DIFF
--- a/rajawali/src/main/java/org/rajawali3d/Object3D.java
+++ b/rajawali/src/main/java/org/rajawali3d/Object3D.java
@@ -372,12 +372,12 @@ public class Object3D extends ATransformable3D implements Comparable<Object3D>, 
 
 	/**
 	 * Positions the object in the plane of the screen/viewport (z = 0), with
-     * x and y specified in GL screen coordinates (origin at bottom left);
-     * Assumes:
-     *  - the object is a direct child of the current scene (not in a container)
-     *  - the current viewport is centered on the Z-axis, and has the given dimensions
-     *  - the current camera is positioned at the given eyeZ
-     *
+	 * x and y specified in GL screen coordinates (origin at bottom left);
+	 * Assumes:
+	 *  - the object is a direct child of the current scene (not in a container)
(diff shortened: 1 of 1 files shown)


rajawali/src/main/java/org/rajawali3d/Object3D.java
```diff
@@ -371,13 +371,18 @@ public boolean isContainer() {
...
 	/**
-	 * Maps screen coordinates to object coordinates
-	 *
-	 * @param x
-	 * @param y
-	 * @param viewportWidth
-	 * @param viewportHeight
-	 * @param eyeZ
+	 * Positions the object in the plane of the screen/viewport (z = 0), with
+     * x and y specified in GL screen coordinates (origin at bottom left);
```
jwoolston: There seems to some white space variation here.

rpicolet: Yep, noticed that too. I'm guessing the "no inconsistent whitespace" guideline in CONTRIBUTING.md means try to maintain whatever formatting style is prevalent in each file?

Also, should I be doing something to squash commits after I push them to my fork? Or should I only ever rewrite history on my local repo?

WIll push the whitespace fix momentarily...

jwoolston: The formatting is all over the place right now. That bit in the guidelines is more about making sure that the commit is not filled with a bunch of noise from whitespace changes. 

As a general rule, once you push something to a public repo, you should not rebase it. We arent super concerned about clean commit histories (look at ours, its a mess) mostly make sure they arent superfluous with a bunch of commits like "changing a constant". Try and get that figured out before submitting the PR. Once you submit a PR, if we ask you to change something then its fine to do it as an additional commit.

rpicolet: Got it. Thanks.

Issue #1561, opened 2015-11-06
Texture coordinates for Plane primitive object not correct


dansoonie:
Currently using the most recent master build(306).
I've been trying to use the plane object for simple testing. And for some reason I'm getting the texture horizontally flipped. After some investigation, I think the problem is related to #1306 and #1533.

I think #1306 on plane facing problem has been solved but resulted in a bug related to #1533. lin1987www points out that his material is reversed. The posted image on the issue suggests that it's the same problem that I am having. The issuer got around the problem by making the plane double sided and rotating along the y axis 180 degrees. However, I don't think that is supposed to be intended.

After doing some testing, I found out that the texture appears as I intend it to when upAxis for the plane is Axis.Y. But when upAxis is Axis.X or Axis.Z, the texture is flipped horizontally.

I didn't go through any detailed analysis but from my observations I believe that there must be some appropriate changes made to the texture coordinates constructing sequence in init() of Plane class.
```
if (mCreateTextureCoords) {
	float u = (float) i / (float) mSegmentsW;
	textureCoords[texCoordCount++] = (1.0f - u) * mNumTextureTiles;
	float v = (float) j / (float) mSegmentsH;
	textureCoords[texCoordCount++] = (1.0f - v) * mNumTextureTiles;
}
```
My guess is that 
```(1.0f - u) * mNumTextureTiles```
should be simply
```u * mNumTextureTiles```
when upAxis is Axis.x or Axis.Z.

Can anyone confirm on this phenomenon? I'm quite surprised that there aren't many people who encountered this problem.

dhelleberg:
Hi,
I had exactly the same issue using a plane for texture mapping and seeing a flipped texture.
Fixed it the same way you did.

I guess this is worth a pull request?

kmare:
Thank you @dansoonie , that fixed it for me (1.0 branch). A push to stable branch is needed. It's indeed weird that no one noticed before.

jwoolston:
> It's indeed weird that no one noticed before.

Its been noticed but there have been PR requests for both sides of the mirror. There seems to be some debate about what is correct.

dansoonie:
@kmare I suspected that somebody had noticed about it and that is probably why I didn't attempt a pull request. I thought something was probably going on and I didn't want to complicate the issue by throwing pull requests at the maintainers.

[...]

PR #1614, opened 2016-03-16
PR for #1607 - combining object color picking and post-processing

### LINKED ISSUE #1607: Combining object color picking and post-processing
[This ticket originated from discussions after #1586 was closed; this post is a partial copy from there.]

...

It's probably better to start by nailing down the basic goals/requirements. I propose that these are:
* to allow all 4 combinations of the two features to inter-operate reliably: default rendering, default rendering plus color picking, post-processing, and post-processing plus color-picking.
* to better isolate the very different "never visible" color-picking GL render details from those of the "visible" renders, and so make it easier to test/maintain both, and to add new features to either
* to minimize changes to the client API for both features,
* within reason, to maximize code re-use and minimize the code footprint

I'll wait to get some feedback on these goals before doing anything more on this. I should say, even if most everyone agrees with these as the goals, I haven't really yet thought of an approach that accomplishes all of them well, so getting some sense of relative priorities would also help me a lot in proposing a design.

...

rpicolet:
Fixes #1607 

## Change Summary
* Scene.java
 * added `doColorPicking()` to isolate color-picking render from normal render() control flow
 * renamed `requestColorPickingTexture()` to `requestColorPicking()` 
* Object3D.java
 * introduced `UNPICKABLE` color/index, made it a default value
 * updated `setPickingColor()`, removed `getPickingColor()`
 * eliminated `mIsPickingEnabled`, now check for `UNPICKABLE`
 * added `renderColorPicking()` for optimized/independent color-picking rendering of the `Object3D` and its children (per #1387)
* ObjectColorPicker.java
 * updated `unregisterObject()` to reset picking color to UNPICKABLE
 * renamed `createColorPickingTexture()` to `pickObject()`
 * in `pickObject()` call to `GLES20.glReadPixels()`, replaced `getDefaultViewportHieght()` with `getViewportHeight()`

## Testing:
* Examples: Object Picking, Touch & Drag now work, and all other examples appear to work same as on master branch
* Picking nested child objects: working on my app...
* Combined color-picking && post-processing: I quickly cut/paste the GreyScale Filter setup into the Object Picking example; monkeys are rendered grey, but seem to have `y = viewportHeight - y`; picking then appears to work as if screen was rendered correctly (e.g. picking the upper left monkey moves the lower left monkey, and vice versa). I can't think of any way the object picking could interfere with the filter renders before anything is even picked, so I think this may be a post-processing-only issue masked by the random nature of the objects in the post-processing examples (which therefore have no natural sense of up/down that would easily reveal the same behavior).

## Concerns
* I was pretty aggressive in optimizing the color-picking rendering methods;
 * setup blending and depth testing just once at the scene level
 * at the object level, eliminated matrix updates, checks/setups for textures, normals, vertex colors, etc.
* I'd really appreciate some more experienced eyes desk-checking my changes and optimizations.  One nice thing here is that if something needs to change, it (probably) _shouldn't_ affect any of the non-color-picking renders
* Haven't really thought about threading/concurrency issues. Probably needs to be e.g. a sync lock on references to `mPickingInfo` in `Scene.requestColorPicking()` and `Scene.render()`?

Any/all questions/suggestions welcome, as usual...

### DIFF
--- a/examples/src/main/java/org/rajawali3d/examples/examples/interactive/TouchAndDragFragment.java
+++ b/examples/src/main/java/org/rajawali3d/examples/examples/interactive/TouchAndDragFragment.java
@@ -14,7 +14,9 @@ import android.widget.TextView;
 import org.rajawali3d.Object3D;
 import org.rajawali3d.examples.R;
 import org.rajawali3d.examples.examples.AExampleFragment;
+import org.rajawali3d.lights.DirectionalLight;
 import org.rajawali3d.materials.Material;
+import org.rajawali3d.materials.methods.DiffuseMethod;
 import org.rajawali3d.math.Matrix4;
 import org.rajawali3d.math.vector.Vector3;
 import org.rajawali3d.primitives.Sphere;
@@ -107,7 +109,14 @@ public class TouchAndDragFragment extends AExampleFragment implements
 			mPicker.setOnObjectPickedListener(this);
 
 			try {
+				DirectionalLight light= new DirectionalLight(-1, 0, -1);
+				light.setPower(1.5f);
+				getCurrentScene().addLight(light);
+				getCurrentCamera().setPosition(0, 0, 4);
+
 				Material material = new Material();
+				material.enableLighting(true);
+				material.setDiffuseMethod(new DiffuseMethod.Lambert());
 
 				for (int i = 0; i < 20; i++) {
 					Sphere sphere = new Sphere(.3f, 12, 12);
--- a/examples/src/main/java/org/rajawali3d/examples/examples/interactive/TouchAndDragFragment.java
+++ b/examples/src/main/java/org/rajawali3d/examples/examples/interactive/TouchAndDragFragment.java
@@ -116,7 +116,7 @@ public class TouchAndDragFragment extends AExampleFragment implements
 					sphere.setX(-4 + (Math.random() * 8));
 					sphere.setY(-4 + (Math.random() * 8));
 					sphere.setZ(-2 + (Math.random() * -6));
-					sphere.setDrawingMode(GLES20.GL_LINE_LOOP);
+					sphere.setDrawingMode(GLES20.GL_TRIANGLES);
 					mPicker.registerObject(sphere);
 					getCurrentScene().addChild(sphere);
 				}
--- a/rajawali/src/main/java/org/rajawali3d/Object3D.java
+++ b/rajawali/src/main/java/org/rajawali3d/Object3D.java
@@ -49,6 +49,8 @@ public class Object3D extends ATransformable3D implements Comparable<Object3D>, 
     public static final int BLUE = 2;
     public static final int ALPHA = 3;
 
+    public static final int UNPICKABLE = -1;
+
 	protected final Matrix4 mMVPMatrix = new Matrix4();
 
 	protected final Matrix4 mMVMatrix = new Matrix4();
@@ -77,9 +79,8 @@ public class Object3D extends ATransformable3D implements Comparable<Object3D>, 
 	protected int mElementsBufferType = GLES20.GL_UNSIGNED_INT;
 
 	protected boolean mIsContainerOnly = true;
-	protected int mPickingColor;
-	protected boolean mIsPickingEnabled = false;
-	protected float[] mPickingColorArray;
+	protected int mPickingIndex;
+	protected float[] mPickingColor;
 
 	protected boolean mFrustumTest = false;
 	protected boolean mIsInFrustum;
@@ -99,6 +100,8 @@ public class Object3D extends ATransformable3D implements Comparable<Object3D>, 
 		mChildren = Collections.synchronizedList(new CopyOnWriteArrayList<Object3D>());
 		mGeometry = new Geometry3D();
 		mColor = new float[] { 0, 1, 0, 1.0f};
+		mPickingColor = new float[4];
+		setPickingColor(UNPICKABLE);
 	}
 
 	public Object3D(String name) {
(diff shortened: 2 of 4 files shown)


rpicolet:
Oh, forgot, BTW. Touch & Drag seems to miss a lot of picks, I think it may be because it's using wireframes for the pickable objects, so the exact touch coordinates may well miss all of the wires...

jwoolston:
Working on this tonight.

jwoolston:
> I think it may be because it's using wireframes for the pickable objects, so the exact touch coordinates may well miss all of the wires

You are correct. If you want to update the examples to render triangles as part of this PR that would be great, otherwise I can do it separately.

rpicolet:
>You are correct. If you want to update the examples to render triangles as part of this PR that would be great, otherwise I can do it separately.

OK, I'll give it a shot...

jwoolston:
Should just be a change from `GLES20.GL_LINE_LOOP` to `GLES20.GL_TRIANGLES`

jwoolston:
@rpicolet can you add or post an example of color picking with post processing. I added greyscale to the object picking as well and I see post processing working, but not object picking, so I am curious what we did differently

jwoolston:
> monkeys are rendered grey, but seem to have y = viewportHeight - y

I believe this is related to the issues with texture coordinates on the screen quad. We should keep this in mind, but leave it to something else to fix. There is already an issue open along these lines: #1561

rpicolet:
> can you add or post an example of color picking with post processing

Hmm, couldn't upload as .java.. Go figure.

Anyway, all I did was cut and paste the lines from GreyScale Filter that create/populate `mEffects`, and then called `mEffects.render()` in `onRender()` instead of `super.onRender()`...

[ObjectPickingFragment.txt](https://github.com/Rajawali/Rajawali/files/177136/ObjectPickingFragment.txt)

jwoolston:
Hmmm, mine looks identical to yours and I cant seem to get it to work (its grey, but no object picking)

rpicolet:
>Should just be a change from GLES20.GL_LINE_LOOP to GLES20.GL_TRIANGLES

That did it, updated the push.

jwoolston:
> Although they now just look like 2D circles... maybe need some lighting change?

Up to you. I think its fine but If you want to add a directional light to the scene that should work fine.

rpicolet:
> Hmmm, mine looks identical to yours and I cant seem to get it to work (its grey, but no object picking)

It's unreliable. I have to poke around a lot to find just the right spots that trigger picking...

jwoolston:
> It's unreliable. I have to poke around a lot to find just the right spots that trigger picking

Hmmm, yes I have noticed the same...seems worth figuring out, though I am not sure if it needs to be fixed here as this does accomplish the initial goal of fixing color picking, including children.

rpicolet:
>Hmmm, yes I have noticed the same...seems worth figuring out, though I am not sure if it needs to be fixed here as this does accomplish the initial goal of fixing color picking, including children.

Maybe we should just see if we can get the problem to occur without color-picking?

Your call, I'm OK either way.

rpicolet:
I just did a quick test, added the first ObjectPIcking  monkey (upper left) to the GreyScale FIlter, and it shows up upside down in the lower left, so that pretty much nails it...

jwoolston:
So further inspection suggests there is some sort of a problem here. I pulled down the entire FrameBuffer  into bitmaps in the debugger and got the following results:

## Normal Color Picking
plain_color_picking

## Color Picking with Greyscale PostProcess
post_processing_picking

The checkerboard in the plain one is to show that the color is transparent (low picking index). As you can see, in the example with post processing, nothing shows in the frame buffer.

rpicolet:
Not at all sure how the framebuffer stuff works, but I'm pretty convinced the problem is not my PR changes.
]

jwoolston:
> but I'm pretty convinced the problem is not my PR change

I agree, they didn't introduce anything. Mostly I want to find out how these two interact. I suspect the fact that more than 1 framebuffer is involved is part of it.

rpicolet:
So, anything else you would like me to do of this? perhaps the sync lock on mPickingInfo? testing other feature combinations?  

BTW, in my app, my touch objects are double-sided transparent panels, so I think that kinda covers a couple more feature integration points...

jwoolston:
Yes, in theory transparency or materials in general should have no effect on this so that is good to hear for confirmation. 

Rather than synchronizing, lets declare it volatile and add a @NonNull annotation to the `requestColorPicking()` method parameter. This saves the overhead of synchronizing while still accomplishing the same result. We don't really care if someone changes the picking info at any time, so long as we are certain every read of it is the most up to date. Adding the non-null annotation defines in the contract that the user cannot set it to null, ensuring it wont become null between our null check and its use. Technically, the user could still do so by breaking the contract, but I am not worried about defensively coding the engine to the point that we start adding extra stuff just to make sure someone doesn't do something against the API.

jwoolston:
once you do that, Ill merge it and track down this issue with the FBOs seperately

rpicolet:
> ...lets declare it volatile and add a @NonNull annotation to the requestColorPicking() method parameter. 

Makes sense. Just because I'm a pedantic pain, I'll point out that there is also a small chance that a second pick could theoretically occur between `doColorPicking(mPickerInfo)` and `mPickerInfo = null` but I seriously doubt the user would be able to tell if it gets ignored so close to the first pick. Will do as you suggest, and will also make ColorPickerInfo immutable just to be a bit safer.

As far as API abuse goes, I would think client apps should really only be coming through the ObjectColorPicker.getObjectAt() method anyway.  Kinda gets at the documentation challenge again.  Maybe a good place to start would be (for some kind, knowledgeable person with lots of time on their hands) to define the supported API calls as interfaces, and just say if ain't in the published IScene or whatever, it ain't supported. Coming someday real soon now!

jwoolston:
> (for some kind, knowledgeable person with lots of time on their hands) to define the supported API calls as interfaces, and just say if ain't in the published IScene or whatever, it ain't supported.

Yes, this is something I have done in a private fork of Rajawali at my employer. It helps considerably with reducing the visibility of API calls for sure. Unfortunately that work is not compatible with merging into the main rajawali repository. It is a fine goal though and I will gladly accept PRs on the issue, provided they are thoughtfully explained of course. I have been making efforts to move in that direction with abstractions of things like ISurfaceRenderer and the like.

PR #1642, opened 2016-03-23
1636 add wear to examples

ToxicBakery:
Wear appears to work, haven't tested the main apk install but it should be good.


ToxicBakery:
Whoops

Issue #1648, opened 2016-03-25
Add .obj and .mtl path support to LoaderObj


ToxicBakery:
#### Rajawali Version or Branch
master (fa1a83d)

#### Device and Android Version
N/A

#### Summary
Add ability to load .obj with .mtl using full path references as an alternative to the model/material renaming currently required.

This was reported and discussed in #1646

#### Steps to Reproduce
N/A

#### Trace or Log Output
N/A

PR #1654, opened 2016-03-29
fix issue #1648

karunass:


rajawali/src/main/java/org/rajawali3d/loader/LoaderOBJ.java
```diff
@@ -94,8 +94,12 @@
     protected final String DIFFUSE_COLOR = "Kd";
     protected final String DIFFUSE_TEX_MAP = "map_Kd";
 
+    private boolean mNeedToRenameMtl = true;
```
jwoolston: Don't set it to `true` here and then `false` in the constructor. Pick one place to initialize it.

karunass: In this case i need to set it "true" in other constructors. Now it is always "true", except this constructor
public LoaderOBJ(Renderer renderer, String fileOnSDCard)

Issue #1663, opened 2016-04-21
gradle tasks target fails with non-obvious error (vuforia) message


contriteobserver:
#### Rajawali Version or Branch

master branch
#### Device and Android Version

linux (ubuntu) build environments, impact on windows builds likely, but not tested
#### Summary

./gradlew tasks fails with non-obvious error message 
#### Steps to Reproduce
- $ git clone https://github.com/Rajawali/Rajawali.git
- $ (cd Rajawali; ANDROID_HOME=~/Android/Sdk ./gradlew tasks)

an error is reported in file '/home/elkq/Rajawali/vuforia/build.gradle' line: 74

Ambiguous method overloading for method java.io.File#<init>.
  Cannot resolve which method to invoke for [null, class java.lang.String] due to overlapping prototypes between:
    [class java.lang.String, class java.lang.String]
    [class java.io.File, class java.lang.String]

What this means is that one has to install the Android NDK and one has to configure the ANDROID_NDK_HOME correctly, so a workaround is:
- $ git clone https://github.com/Rajawali/Rajawali.git
- $ (cd Rajawali; ANDROID_HOME=~/Android/Sdk ANDROID_NDK_HOME=~/Android/Sdk/ndk-bundle ./gradlew tasks)

Least surprising behavior is that the tasks and *WithoutVuforia* targets should not depend upon the NDK. Similarily, *WithVuforia* targets should issue an error message like the one issued when one forgets to set ANDROID_HOME.

jwoolston:
Closing this issue as resolved via #1664

PR #1664, opened 2016-04-22
added an unknown NDK location error message.

contriteobserver:
Implemented the second part of of issue 1663, added an error message if NDK is not found.
Do not know the vuforia build components well enough to risk messing with that.


ToxicBakery:
Adding issue reference #1663

PR #1724, opened 2016-07-26
fixed inconsistency in obtaining 1/w

GXmachin:
There were some inconsistency in calculating the inverse of w used for perspective division after multiplying the vector and the matrix. 

The same for projectAndCreateVector method aswell


ToxicBakery:
@jwoolston

jwoolston:
Merging to take into account with my coordinate system checks.

PR #1726, opened 2016-07-28
Changes in support of a new material plugin

jwoolston:
While working on a material plugin, I encountered some issues I wished to correct:

- `Geometry3D` now has a new method, `addBuffer` which allows arbitrary VBOs to be added with a String key, and they will automatically be managed by `Geometry3D` just like the standard buffers. With this change, the buffers must be added to the `BufferInfo` object prior to being added.

- `Material#createShaders()` is not `protected` access rather than `private`.
- `Material#onPreVertexShaderInitialize(VertexShader)` and `Material#onPreFragmentShaderInitialize(FragmentShader)` methods have been added with default empty implementations.

- `AShader` has been modified to fetch preprocessor directives from all added plugins which are instances of `AShader` themselves.

- Support for non-uint index arrays has been removed. Some uncertainty on the validity of this decision for wear devices remains.

- Exporter functions have been deleted completely as they have long been deprecated.

- Version numbers have been bumped to 1.2.x to reflect the changes.


rajawali/src/main/java/org/rajawali3d/Geometry3D.java
```diff
@@ -48,658 +50,607 @@
...
+     * @see VertexAnimationObject3D
+     */
+    public void setData(BufferInfo vertexBufferInfo, BufferInfo normalBufferInfo,
+                        float[] textureCoords, float[] colors, int[] indices, boolean createVBOs) {
+        if (textureCoords == null || textureCoords.length == 0) {
+            textureCoords = new float[(mNumVertices / 3) * 2];
+        }
+        setTextureCoords(textureCoords);
+        if (colors == null || colors.length == 0) {
+            setColors(0xff000000 + (int) (Math.random() * 0xffffff));
```
ToxicBakery: It might be better to make these constants `OPAQUE` and `WHITE` then you can simply do 

```
OPAQUE | (int) (Math.random() * WHITE);
```

Thinking about that more, I'm not sure what the value is in having a huge range of color possibilities.

```
public static final Random RANDOM = new Random();

...

// Create a 'random' color from a a set of 6 possible color shades of RGB then set it opaque 
OPAQUE | 0xf << RANDOM.nextInt(7);
```

ToxicBakery:
Looks good to me. I agree with drop the exporter functionality.

ToxicBakery:
👍

Issue #1760, opened 2016-09-03
unexpected behaviour from Geometry3D.addFromGeometry3D


contriteobserver:
#### Rajawali Version or Branch
master

#### Device and Android Version
SM-T230N

#### Summary
`NullPointerException` thrown in `Geometry3D.addFromGeometry3D`

#### Steps to Reproduce
Was hoping to use `Geometry3D.addFromGeometry3D` to augment geometry at runtime.

```
        protected void initScene() {
            getCurrentScene().setBackgroundColor(Color.BLUE);

            try {
                Geometry3D geometry;
                Material red = new Material();
                red.setColor(Color.RED);
                Material cyan = new Material();
                cyan.setColor(Color.CYAN);
                Material yellow = new Material();
                yellow.setColor(Color.YELLOW);
                Material magenta = new Material();
                magenta.setColor(Color.MAGENTA);

                float[] vertices = new float[] {
                        -0.5f, 0.5f, 0,
                        -0.5f,-0.5f, 0,
                        0.5f,-0.5f, 0,
                        0.5f, 0.5f, 0,
                };
                float[] normals = null;
                float[] textureCoords = null;
                float[] colors = null;
                int[] indices = new int[] {
                        0,1,2,
                        0,2,3,
                };
                boolean createVBOs = true;

                geometry = new Geometry3D();
                geometry.setData(vertices, normals, textureCoords, colors, indices, createVBOs);
                Object3D obj1 = new Object3D();
                obj1.isContainer(false);
                obj1.getGeometry().setData(vertices, normals, textureCoords, colors, indices, createVBOs);
                obj1.setPosition(-0.5,0.5,0);
                obj1.setMaterial(red);
                getCurrentScene().addChild(obj1);

                geometry = new Geometry3D();
                geometry.setData(vertices, normals, textureCoords, colors, indices, createVBOs);
                Object3D obj2 = new Object3D();
                obj2.isContainer(false);
                obj2.getGeometry().setVertices(vertices);
                obj2.getGeometry().setIndices(indices);
                obj2.getGeometry().createBuffers();
                obj2.setPosition(0.5,0.5,0);
                obj2.setMaterial(cyan);
                getCurrentScene().addChild(obj2);

                geometry = new Geometry3D();
                geometry.setData(vertices, normals, textureCoords, colors, indices, createVBOs);
                Object3D obj3 = new Object3D();
                obj3.isContainer(false);
[...]

PR #1761, opened 2016-09-04
Fix for unexpected behaviour from Geometry3D.addFromGeometry3D

contriteobserver:
Proposed Fix for #1760 

Augmented getFloatArrayFromBuffer() and getIntArrayFromBuffer() to return null when the input buffer is null.

Augmented addFromGeometry3D() to test for null array values.

### DIFF
--- a/rajawali/src/main/java/org/rajawali3d/Geometry3D.java
+++ b/rajawali/src/main/java/org/rajawali3d/Geometry3D.java
@@ -194,30 +194,34 @@ public class Geometry3D {
 
     public static float[] getFloatArrayFromBuffer(FloatBuffer buffer) {
         float[] array = null;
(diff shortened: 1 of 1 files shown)


jwoolston:
Looking at the referenced issue and the stack trace, it looks to me that the Normal buffer was null, and I find this concerning. While not annotated as such, the intention of the `*ArrayFromBuffer` methods is that the input buffer not be null. Would you be opposed to altering this PR to instead check that the buffer is not null prior to calling these methods?

contriteobserver:
Submitted pull request #1762 to check that the buffer is not null prior to calling `*ArrayFromBuffer` methods. Feel free to pull in whichever version fits better in the larger context.

PR #1762, opened 2016-09-05
Another approach to fixing unexpected behaviour from Geometry3D.addFromGeometry3D

contriteobserver:
Another approach to fixing #1760 

Augmented addFromGeometry3D() to test for null array values.

### DIFF
--- a/rajawali/src/main/java/org/rajawali3d/Geometry3D.java
+++ b/rajawali/src/main/java/org/rajawali3d/Geometry3D.java
@@ -263,22 +263,28 @@ public class Geometry3D {
         float[] newColors = null;
         float[] newTextureCoords = null;
         int[] newIntIndices = null;
-        float[] mVerticesArray = null;
-        float[] mNormalsArray = null;
-        float[] mColorsArray = null;
-        float[] mTextureCoordsArray = null;
-        int[] mIndicesArray = null;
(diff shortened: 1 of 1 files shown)


contriteobserver:
A variant of pull request #1761

rajawali/src/main/java/org/rajawali3d/Geometry3D.java
```diff
@@ -263,22 +263,28 @@ public void addFromGeometry3D(Vector3 offset, Geometry3D geometry, boolean creat
         float[] newColors = null;
         float[] newTextureCoords = null;
         int[] newIntIndices = null;
-        float[] mVerticesArray = null;
-        float[] mNormalsArray = null;
-        float[] mColorsArray = null;
-        float[] mTextureCoordsArray = null;
-        int[] mIndicesArray = null;
+        float[] mVerticesArray = new float[0];
```
jwoolston: What was the purpose for initializing these values?

contriteobserver: We have to initialise these values to prevent calls to `concatAllFloat()` failing with NullPointerExceptions.

jwoolston:
After seeing the differences, I think the first approach is cleaner. It has been merged. Sorry for the trouble and thankyou for the contribution.

Issue #1766, opened 2016-09-08
unexpected behaviour when repeatedly calling Geometry3D.addFromGeometry3D()


contriteobserver:
#### Rajawali Version or Branch
master

#### Device and Android Version
SM-T230N

#### Summary
When repeatedly calling `Geometry3D.addFromGeometry3D()` a `java.nio.BufferOverflowException` is thrown. Only possible to run into this issue after fix for #1760 was implemented.

#### Steps to Reproduce
```
        protected void initScene() {
            getCurrentScene().setBackgroundColor(Color.BLUE);

            try {
                Geometry3D geometry;
                Material yellow = new Material();
                yellow.setColor(Color.YELLOW);

                float[] vertices = new float[] {
                        -0.5f, 0.5f, 0,
                        -0.5f,-0.5f, 0,
                        0.5f,-0.5f, 0,
                        0.5f, 0.5f, 0,
                };
                float[] normals = null;
                float[] textureCoords = null;
                float[] colors = null;
                int[] indices = new int[] {
                        0,1,2,
                        0,2,3,
                };
                boolean createVBOs = false;

                geometry = new Geometry3D();
                geometry.setData(vertices, normals, textureCoords, colors, indices, createVBOs);

                Object3D obj = new Object3D();
                obj.isContainer(false);
                obj.getGeometry().addFromGeometry3D(new Vector3(-1,-1,-1), geometry, createVBOs);
                obj.getGeometry().addFromGeometry3D(new Vector3( 0, 0, 0), geometry, createVBOs);
                obj.getGeometry().addFromGeometry3D(new Vector3( 1, 1, 1), geometry, createVBOs);
                obj.getGeometry().createBuffers();
                obj.setMaterial(yellow);
                getCurrentScene().addChild(obj);

                getCurrentCamera().setLookAt(obj.getPosition());
            } catch (Exception e) {
                e.printStackTrace();
            }
        }
```

#### Trace or Log Output
```
09-07 16:23:12.395 31969-31999/com.example.geometryfail W/System.err: java.nio.BufferOverflowException
09-07 16:23:12.395 31969-31999/com.example.geometryfail W/System.err:     at java.nio.Buffer.checkPutBounds(Buffer.java:189)
09-07 16:23:12.395 31969-31999/com.example.geometryfail W/System.err:     at java.nio.DirectByteBuffer.put(DirectByteBuffer.java:326)
[...]

PR #1767, opened 2016-09-08
fix for unexpected behaviour when repeatedly calling Geometry3D.addFromGeometry3D()

contriteobserver:
Fix for #1766 

`setVertices()` already had a boolean override, this change adds a similar ability to

- `setNormals()`
- `setTextureCoords()`
- `setColors()`
- `setIndices()`

and their associated buffers


jwoolston:
Thanks for the fix!

PR #1768, opened 2016-09-08
Updates renderer startup message and adds a script check for PR to pr…

jwoolston:
…event attempting to decrypt signing key on PRs.

Signed-off-by: Jared Woolston <jwoolston@tenkiv.com>


jwoolston:
DO NOT MERGE

PR #1769, opened 2016-09-09
Do not remove item from array to maintain the consistency with the index

duiker101:
When calling unregisterObject  the index doesn't change but now the size of the object lookup array is smaller and it might happen that some objects are no longer clickable or return the wrong one so instead of removing the item from the array we can just set the value to null.


jwoolston:
Thanks for the bug fix!

PR #1770, opened 2016-09-11
Get the first empty in the loader threads SparseArray and use it as ID for the new thread.

duiker101:
For the loading new models asynchronously the id of the new thread is just the size of the SparseArray at the moment, the id is then used to retrieve the the thread at the end of the execution but if other threads finished before they might influence the array leading to a NullPointerException when trying to retrieve the thread.

A more simple solution could be to have the last ID as a member but I wouldn't want to run into max int complications if there are A LOT of models loaded constantly and it shouldn't affect performance.


jwoolston:
This solution appears to suffer from the same thread safety problem as the current implementation, mainly because `SparseArray` is not thread safe and there is no locking. I think the best, most straight forward approach here would be to add a final `AtomicInteger` to `Renderer` and use `incrementAndGet()` to fetch the new id.

jwoolston:
Looks good. Thanks :+1:

Issue #1774, opened 2016-09-16
Object3d.getBoundingBox() returning null. Calling java.nio package methods which shouldn't be called


JehandadK:
#### Rajawali Version or Branch

stable release org.rajawali3d:rajawali:1.0.325@aar
#### Device and Android Version

Google Tango Device API 19 Release Vega
#### Summary

Cannot render a bounding box around a 3dObject loaded from obj file. While trying getBoundingBox a Nullpointer is thrown
#### Steps to Reproduce

```
        Log.d(TAG, mObject.getGeometry() + "");
        Log.d(TAG, mObject.getGeometry().getBoundingBox() + ""); // Error thrown here
        mObject.setShowBoundingVolume(true);
```
#### Trace or Log Output

```
java.lang.NullPointerException: Attempt to invoke virtual method 'java.nio.Buffer java.nio.FloatBuffer.rewind()' on a null object reference
                      at org.rajawali3d.bounds.BoundingBox.calculateBounds(BoundingBox.java:143)
                      at org.rajawali3d.bounds.BoundingBox.<init>(BoundingBox.java:73)
                      at org.rajawali3d.Geometry3D.getBoundingBox(Geometry3D.java:1011)
                      at com.royalcyber.ontango.SceneRenderer.onModelLoadComplete(SceneRenderer.java:110)
                      at org.rajawali3d.renderer.RajawaliRenderer$19.handleMessage(RajawaliRenderer.java:1046)
                      at android.os.Handler.dispatchMessage(Handler.java:102)
                      at android.os.Looper.loop(Looper.java:136)
                      at android.app.ActivityThread.main(ActivityThread.java:5030)
                      at java.lang.reflect.Method.invoke(Native Method)
                      at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:793)
                      at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:609)
```

Related to: http://stackoverflow.com/questions/39522747/rendering-a-frame-around-a-3d-object-using-opengl/39523907?noredirect=1#comment66363414_39523907

jwoolston:
This crash happened because the geometry you are trying to calculate a bounding box for does not appear to have a vertex buffer.

JehandadK:
How can I create it?

jwoolston:
Is this a child object? If so, it is expected, you will need to get the bounding box from the parent.

JehandadK:
[...]

PR #1780, opened 2016-09-28
Enable parent objects with no vertices to create Bounding Box

JehandadK:
Fix for issue #1774 

Opened pull request to get some guidance. For now only refactored code that search for min max points in Object3D's vertices.

### DIFF
--- a/rajawali/src/main/java/org/rajawali3d/Geometry3D.java
+++ b/rajawali/src/main/java/org/rajawali3d/Geometry3D.java
@@ -1074,4 +1074,8 @@ public class Geometry3D {
     public void setBuffersCreated(boolean created) {
         mHaveCreatedBuffers = created;
     }
+
+    void setBoundingBox(BoundingBox boundingBox){
+        this.mBoundingBox = boundingBox;
+    }
 }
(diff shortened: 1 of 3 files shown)


JehandadK:
The BoundingBox is a part of Geometry3D and not a part of Object3D, but creating a bounding box for an Object3D with no vertices requires access to its children. Should I consider 

- moving the bounding box to Object3D
OR
- give a public/package local setter for BoundingBox in Geometry3D.

jwoolston:
> Should I consider
> 
> moving the bounding box to Object3D OR
> give a public/package local setter for BoundingBox in Geometry3D

Go with the setter on `Geometry3D`. Keep the visibility as low as possible.

JehandadK:
Object3D can have the following cases

- It has vertices, in which case the bounding box was created fine
- It has no vertices but has children, and now it will create a bounding box correctly
- It has vertices as well as children. What should we do in this case? I dont know if this is valid opengl scenario.
- It has no vertices and no children, in which case getting the bounding box based on Maximum and minimum Double Values. I suggest returning a default Bounding Box in this case

jwoolston:
> It has vertices as well as children. What should we do in this case? I dont know if this is valid opengl scenario.

This is valid, but somewhat undefined in the current Rajawali. I am working to clarify that moving forward. In the meantime, the bounding box should be based on both the object verticies and its children.



> It has no vertices and no children, in which case getting the bounding box based on Maximum and minimum Double Values. I suggest returning a default Bounding Box in this case

For various reasons, a box based on the min/max double values is appropriate so that the box essentially always intersects.

JehandadK:
Tests working fine on my own app as well as examples.
device-2016-09-30-170122
device-2016-09-30-165759

jwoolston:
Looks good to me. Thanks for taking this on.

Issue #1784, opened 2016-10-03
Can't render Object3D with zero position and scale


rpicolet:
#### Rajawali Version or Branch
1.1.610

#### Summary
Rendering `Object3D` with position and scale both 0 results in `IllegalStateException` from `Matrix4.inverse()`.  This worked in older versions, prior to the introduction of the `throws` clause on `Matrix4.inverse()`.  (The use case is adding an object to the scene which is to be made visible via a scaling animation.)

Not sure what the best/preferred fix is (handle the exception in setToNormalMatrix()?; remove the throws?; add a check for isInvertible()?), but I'm willing to take direction to create a PR for it.

#### Steps to Reproduce
Set position and scale of an `Object3D` to zero; next frame render fails.

#### Trace or Log Output
10-03 15:10:35.580 637-687/com.stupidhardgames.stuawki E/AndroidRuntime: FATAL EXCEPTION: GLThread 5017
                                                                         Process: com.stupidhardgames.stuawki, PID: 637
                                                                         java.lang.IllegalStateException: Matrix is singular and cannot be inverted.
                                                                             at org.rajawali3d.math.Matrix4.inverse(Matrix4.java:348)
                                                                             at org.rajawali3d.math.Matrix4.setToNormalMatrix(Matrix4.java:726)
                                                                             at org.rajawali3d.materials.Material.setModelMatrix(Material.java:1076)
                                                                             at org.rajawali3d.Object3D.render(Object3D.java:300)
                                                                             at org.rajawali3d.Object3D.render(Object3D.java:343)
                                                                             at org.rajawali3d.Object3D.render(Object3D.java:343)
                                                                             at org.rajawali3d.Object3D.render(Object3D.java:343)
                                                                             at org.rajawali3d.Object3D.render(Object3D.java:343)
                                                                             at org.rajawali3d.Object3D.render(Object3D.java:343)
                                                                             at org.rajawali3d.Object3D.render(Object3D.java:343)
                                                                             at org.rajawali3d.Object3D.render(Object3D.java:343)
                                                                             at org.rajawali3d.Object3D.render(Object3D.java:343)
                                                                             at org.rajawali3d.Object3D.render(Object3D.java:343)
                                                                             at org.rajawali3d.Object3D.render(Object3D.java:343)
                                                                             at org.rajawali3d.Object3D.render(Object3D.java:184)
                                                                             at org.rajawali3d.scene.Scene.render(Scene.java:1135)
[...]

PR #1786, opened 2016-10-04
Addressing issue #1784:

rpicolet:
- adding throws IllegalStateException to Matrix4.setToNormal() as it depends on Matrix4.inverse()
- putting call to setToNormal() in Material.setModelMatrix() in try/catch block, thus enabling continued rendering of degenerate/zero-scale Object3Ds (which have singular model matrices)

Tested on existing app where issue surfaced, no changes required.

### DIFF
--- a/rajawali/src/main/java/org/rajawali3d/materials/Material.java
+++ b/rajawali/src/main/java/org/rajawali3d/materials/Material.java
@@ -15,6 +15,7 @@ package org.rajawali3d.materials;
 import android.graphics.Color;
 import android.opengl.GLES20;
 import android.support.annotation.NonNull;
+import android.util.Log;
 import org.rajawali3d.BufferInfo;
 import org.rajawali3d.Object3D;
 import org.rajawali3d.lights.ALight;
@@ -1077,7 +1078,7 @@ public class Material {
         try {
             mNormalMatrix.setToNormalMatrix();
         } catch (IllegalStateException exception) {
-            // modelMatrix is degenerate (zero scale)
+            RajLog.d("modelMatrix is degenerate (zero scale)...");
         }
         float[] matrix = mNormalMatrix.getFloatValues();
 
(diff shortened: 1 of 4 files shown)


rajawali/src/main/java/org/rajawali3d/materials/Material.java
```diff
@@ -1073,7 +1073,12 @@ public void setModelMatrix(Matrix4 modelMatrix) {
         mModelMatrix = modelMatrix;//.getFloatValues();
         mVertexShader.setModelMatrix(mModelMatrix);
 
-        mNormalMatrix.setAll(modelMatrix).setToNormalMatrix();
+        mNormalMatrix.setAll(modelMatrix);
+        try {
+            mNormalMatrix.setToNormalMatrix();
+        } catch (IllegalStateException exception) {
+            // modelMatrix is degenerate (zero scale)
```
jwoolston: Lets print a message here. Given the use case, I think a `RajLog.d()` is fitting so it only surfaces if debug output is enabled.

rpicolet: Or, rather than computing the determinant and getting the exception, perhaps we could just check for zero scale before even calling setToNormal()?

rpicolet: Or maybe even just skip the render altogether for zero-scale objects. like we do for invisibles?

jwoolston: If it is done as a `Rajlog.d`, it will only show if `RajLog.setDebugEnabled(true)` is called, in which case the logs are going to be quite verbose anyway as all the materials will be printing their shader source, uniform handles, etc. The goal is just to give some indication to the developer, not the user, of what is happening, particularly when it happens for other reasons rather than the intentional use case you present. Additionally, it would not be bad to force the rest of the render process to be skipped.

jwoolston: For something like this, with we prefer an additional commit. The only time we want things rebased would be if the initial PR exceeds a few commits. For example, when I PR the 2.0 changes, there will likely be hundreds of commits. I will rebase this into 1.

jwoolston:
@rpicolet Thanks for fixing this. I will push out a new release soon.

rpicolet:
Cool. No hurry for the release on my part.

Issue #1796, opened 2016-10-14
unexpected behaviour from Vignette Pass


contriteobserver:
#### Rajawali Version or Branch
master

#### Device and Android Version
SM-T230N

#### Summary
vignette pass sometimes reduces the brightness of the center of the image, instead of the edges as expected.
![screenshot-2016-10-13](https://cloud.githubusercontent.com/assets/17471201/19378310/b9759670-91a0-11e6-841a-6bcc933a0abf.png)

#### Steps to Reproduce
```
        @Override
        protected void initScene() {
            try {
                getCurrentScene().setBackgroundColor(Color.BLUE);
                DirectionalLight key = new DirectionalLight(-2,-2,-2);
                key.setPosition(new Vector3(2,2,2));
                key.setPower(2);
                getCurrentScene().addLight(key);

                Material material = new Material();
                material.enableLighting(true);
                material.setDiffuseMethod(new DiffuseMethod.Lambert());
                material.setAmbientColor(Color.BLUE);
                Object3D obj = new Cube(1);
                obj.setMaterial(material);
                getCurrentScene().addChild(obj);

                Animation3D anim = new RotateOnAxisAnimation(new Vector3(1,1,1), 360);
                anim.setTransformable3D(obj);
                anim.setDurationMilliseconds(5678);
                anim.setRepeatMode(Animation.RepeatMode.INFINITE);
                anim.play();

                getCurrentScene().registerAnimation(anim);

                mEffects = new PostProcessingManager(this);

                RenderPass renderPass = new RenderPass(getCurrentScene(), getCurrentCamera(), 0);
                mEffects.addPass(renderPass);

                EffectPass vignettePass = new VignettePass(0.75f,0.25f,0.50f,mCurrentViewportWidth, mCurrentViewportHeight);
                vignettePass.setRenderToScreen(true);
                mEffects.addPass(vignettePass);

            } catch (Exception e) {
                Log.d(getLocalClassName(), e.toString());
            }
        }

        @Override
        protected void onRender(long ellapsedRealtime, double deltaTime) {
            //super.onRender(ellapsedRealtime, deltaTime);
            mEffects.render(ellapsedRealtime, deltaTime);
        }
```

jwoolston:
Resolved via #1797

PR #1797, opened 2016-10-14
Fix for issue #1796 unexpected behaviour from Vignette Pass

contriteobserver:
Inner and Outer radii are now correctly derived from uRadius and uSmoothness.

Turns out that when vignette pass worked, it was due to undefined behaviour of smoothstep.
(smoothstep results are undefined if edge0 ≥ edge1)


jwoolston:
Thanks for this. I had noticed it but never put the time to look into it so you've saved me a lot of effort here.

PR #1800, opened 2016-10-22
Set back texture buffer position to zero

jguomoto:
Otherwise if we call setTextureCoords() twice consecutively will cause a `BufferOverflowException`.


jwoolston:
I am declining this PR as it is functioning as intended. If your wish is to append buffers, that is what the override flag is for. You are expected to create an appended array, then call `setTextureCoords(float[], override)` with the new longer array and the `override` flag set to `true`.

jwoolston:
@jguomoto All the same, thank-you for your interest in helping out.

jguomoto:
ah! makes sense, thanks. I am currently using a old version which doesn't have the override param, I guess I will update my library then.

PR #1802, opened 2016-10-27
WIP: Proposed changes to render structure

rpicolet:
Review-only, do not pull/merge.

Not sure how much to write here.  Created a v2.0-engine branch, since everything here is engine-framework stuff; no multipass yet.

Fairly aggressive with refactoring/re-naming, trying to promote a clear distinction between frame-management and rendering views, and between models and views.

All feedback welcome!


rajawali/src/main/java/c/org/rajawali3d/engine/EngineImpl.java
```diff
@@ -220,105 +244,104 @@ public void onRenderSurfaceCreated(EGLConfig config, GL10 gl, int width, int hei
...
-     * @param deltaTime        {@code double} Time passed since last frame, in seconds.
-     */
-    protected void render(final long ellapsedRealtime, final double deltaTime) throws InterruptedException {
-        if (currentRenderable != null) {
-            currentRenderable.render(ellapsedRealtime, deltaTime);
-        } else {
-            Log.w(TAG, "No renderable has been set!");
+    protected void endFrame(double deltaTime) {
+        try {
+            // TODO not sure if order matters here, just swapping from startFrame() on principle...
```
jwoolston: I imagine this is something that will become apparent as the implementations are fleshed out, but I'm inclined to suspect that the order is backwards. From my understanding, I am thinking we will want `RenderView`s to receive the callback prior to the `RenderModel`s. Funny side joke about that: [Flesh vs Flush](http://theoatmeal.com/comics/flesh_vs_flush)

rajawali/src/main/java/c/org/rajawali3d/engine/EngineImpl.java
```diff
@@ -220,105 +244,104 @@ public void onRenderSurfaceCreated(EGLConfig config, GL10 gl, int width, int hei
...
+            for (RenderView renderView : renderViews) {
```
jwoolston: More `for-each` loops.

rajawali/src/main/java/c/org/rajawali3d/engine/EngineImpl.java
```diff
@@ -220,105 +244,104 @@ public void onRenderSurfaceCreated(EGLConfig config, GL10 gl, int width, int hei
...
-     * @param nextRenderable The {@link Renderable} to switch to.
-     */
-    @GuardedBy("nextRenderableLock")
-    @GLThread
-    void switchRenderable(@NonNull Renderable nextRenderable) {
-        RajLog.d("Switching from renderable: " + currentRenderable + " to renderable: " + nextRenderable);
-        currentRenderable = nextRenderable;
-        currentRenderable.restoreForNewContextIfNeeded();
+    @Override
+    public void onOffsetsChanged(float xOffset, float yOffset, float xOffsetStep, float yOffsetStep, int xPixelOffset, int yPixelOffset) {
```
jwoolston: We need to create an interface and add a `@Nullable` delegate member to this class to propagate this and `onTouchEvent(MotionEvent)` calls to optional user code since we are removing the extension of what used to be `Renderer` classes.

rajawali/src/main/java/c/org/rajawali3d/engine/RenderModel.java
```diff
@@ -0,0 +1,44 @@
+package c.org.rajawali3d.engine;
+
+import android.support.annotation.Nullable;
+import c.org.rajawali3d.annotations.GLThread;
+
+/**
+ * Interface to be implemented by models that are to be displayed in RednerViews managed by the Engine
+ *
+ * Author: Randy Picolet
```
jwoolston: Use the javadoc `@author` label please. It results in nice formatting by the documentation processors.

rajawali/src/main/java/c/org/rajawali3d/engine/RenderModel.java
```diff
@@ -0,0 +1,44 @@
+package c.org.rajawali3d.engine;
+
+import android.support.annotation.Nullable;
+import c.org.rajawali3d.annotations.GLThread;
+
+/**
+ * Interface to be implemented by models that are to be displayed in RednerViews managed by the Engine
```
jwoolston: I'm trying to get into the habit of using {@link SOME CLASS} markups in my javadoc - it would be a good habit for the entire engine.

rajawali/src/main/java/c/org/rajawali3d/engine/RenderView.java
```diff
@@ -0,0 +1,66 @@
...
+    void setEngine(@Nullable Engine engine);
+
+    /**
+     *
+     * @return the {@link RenderModel} instance presented in this view, if any
+     */
+    @Nullable
+    RenderModel getRenderModel();
+
+    void setViewport(Rect rect);
```
jwoolston: Can this ever be null? We should use the appropriate annotation here regardless of the answer (`@NonNull` or `@Nullable`)

rajawali/src/main/java/c/org/rajawali3d/engine/RenderView.java
```diff
@@ -0,0 +1,66 @@
...
+
+    Rect getViewport();
```
jwoolston: Same question as `setViewport()`

rajawali/src/main/java/c/org/rajawali3d/engine/RenderView.java
```diff
@@ -0,0 +1,66 @@
...
+import android.support.annotation.Nullable;
+import c.org.rajawali3d.annotations.GLThread;
+
+/**
+ * Interface to be implemented by views that are to be managed by the Engine. A RenderView represents
+ * a viewport on the render surface
+ *
+ * @author Jared Woolston (Jared.Woolston@gmail.com)
+ */
+public interface RenderView {
```
jwoolston: I'm thinking it makes sense to have `RenderView` implement `SurfaceCallback` or otherwise be notified of the surface size changed (such as through a method independent of `SurfaceCallback` that is part of the `RenderView` interface). This would allow `RenderView` to have enough information to use relative coordinates for its size and automatically correct its viewport settings for changes.

rajawali/src/main/java/c/org/rajawali3d/scene/SceneFrameCallback.java
```diff
@@ -18,51 +19,32 @@
...
-
-    /**
-     * Post frame render callback. Called after all frame drawing has completed, including plugins.
-     *
-     * @param sceneTime {@code long} Rendering elapsed time in nanoseconds.
-     * @param deltaTime {@code double} Time passed since last frame in seconds.
-     */
-    @GLThread
-    void onPostFrame(long sceneTime, double deltaTime);
+    void onFrameEnd(double deltaTime);
```
jwoolston: Similar to above, I think we need to keep the elapsed time here.

rajawali/src/main/java/c/org/rajawali3d/sceneview/SceneView.java
```diff
@@ -0,0 +1,227 @@
...
+import c.org.rajawali3d.scene.Scene;
+import c.org.rajawali3d.scene.graph.NodeMember;
+import java.util.List;
+import java.util.concurrent.locks.Lock;
+import net.jcip.annotations.GuardedBy;
+import org.rajawali3d.math.Matrix4;
+import org.rajawali3d.util.RajLog;
+
+/**
+ * Author: Randy Picolet
```
jwoolston: `@author` tag.

rajawali/src/main/java/c/org/rajawali3d/sceneview/SceneView.java
```diff
@@ -0,0 +1,227 @@
...
+    private Camera nextCamera; // The camera the scene should switch to on the next frame.
+    private final Object nextCameraLock = new Object(); // Camera switching lock
+
+    @NonNull
+    @GLThread
+    private Matrix4 viewProjectionMatrix = new Matrix4();
+
+    @Nullable
+    @GLThread
+    private ObjectRenderer lastUsedRenderer; // Reference to the last used object engine
```
jwoolston: I had it wrong before, but this is something that needs to be held at the `Engine` level. It is relevant to to the GL context as a whole.

rajawali/src/main/java/c/org/rajawali3d/sceneview/SceneView.java
```diff
@@ -0,0 +1,227 @@
...
+
+    @Nullable
+    Lock currentlyHeldWriteLock;
+    @Nullable
+    Lock currentlyHeldReadLock;
+
+    // TODO Not sure about volatile semantics here
+    private volatile boolean isDisabled;
```
jwoolston: This is a perfect use of `volatile`

rajawali/src/main/java/c/org/rajawali3d/sceneview/SceneView.java
```diff
@@ -0,0 +1,227 @@
...
+public class SceneView implements RenderView {
+    private static final String TAG = "SceneView";
+
+    @Nullable
+    private Engine engine;
+
+    @NonNull
+    private Scene scene;
+
+    // TODO probably need a decorator Viewport class around the Rect to flag changes...
```
jwoolston: Yes...this is an interesting idea. We should discuss in more detail how the viewport dimensions should work and what is responsible for managing them.

rajawali/src/main/java/c/org/rajawali3d/sceneview/SceneView.java
```diff
@@ -0,0 +1,227 @@
...
+    public SceneView(@NonNull Rect viewport) {
+        this.viewport = viewport;
+    }
+
+    public SceneView(@NonNull Scene scene) {
+        this.viewport = new Rect();
+        this.scene = scene;
+    }
+
+    public SceneView(@NonNull Rect viewport, @NonNull Scene scene) {
```
jwoolston: With the class members `viewport` and `scene` being `@NonNull`, this should be the only constructor. Remove the other two. Another alternative is to set the value of `this.viewport` in the declaration and in the constructors do `this.viewport.set(viewport)` or whatever the actual syntax is.

rajawali/src/main/java/c/org/rajawali3d/sceneview/SceneView.java
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    @Override
+    public void isDisbled(boolean disabled) {
+        this.isDisabled = disabled;
+    }
+
+    @Override
+    public boolean isDisabled() {
+        return false;
```
jwoolston: Why not return `isDisabled`? (You used Android Studio's auto implement and forgot to change it didn't you)

rajawali/src/main/java/c/org/rajawali3d/sceneview/SceneView.java
```diff
@@ -0,0 +1,227 @@
...
+            if (currentlyHeldReadLock != null) {
+                currentlyHeldReadLock.unlock();
+            }
+        }
+    }
+
+    @GLThread
+    @Override
+    public void onRenderView() throws InterruptedException {
+        // TODO Lock should probably be on the whole Scene, not just its Graph
```
jwoolston: The lock was intentionally placed on only the graph to allow for more granular thread control. The expectation is that other things which need protection would be protected separately. The way the scene graph is structured, there should be no interaction with any other data models in the engine.

rajawali/src/main/java/c/org/rajawali3d/sceneview/SceneView.java
```diff
@@ -0,0 +1,227 @@
...
+                nextCamera = null;
+            }
+        }
+
+        // Prepare the camera matrices
+        final Matrix4 viewMatrix = currentCamera.getViewMatrix();
+        final Matrix4 projectionMatrix = currentCamera.getProjectionMatrix();
+
+        if (projectionMatrix == null) {
+            throw new IllegalStateException("Cannot render while current camera has a null projection matrix.");
```
jwoolston: This exception was wrong in my initial code and I have since changed it to an error log and break out. We will want to update this to be in line with that as well.

rajawali/src/main/java/c/org/rajawali3d/sceneview/SceneView.java
```diff
@@ -0,0 +1,227 @@
...
+
+        // still TODO: This will be an interaction point with the render pass manager. We don't want to check the
+        // intersection with the camera multiple times. One possible exception would be for shadow mapping. Probably
+        // a loop
+
+        // Fetch the current render type
+        int type = 0;
+
+        // Loop each node and draw
+        for (NodeMember member : intersectedNodes) {
```
jwoolston: I know I created this `for-each` loop originally. I want to remove it.

rajawali/src/main/java/c/org/rajawali3d/sceneview/SceneView.java
```diff
@@ -0,0 +1,227 @@
...
+    public void onFrameEnd(double deltaTime) throws InterruptedException {
+        // TODO Anything?
+    }
+
+    @RequiresReadLock
+    @GLThread
+    protected void internalRender() throws IllegalStateException {
+
+        // TODO Probably need to mimic the nextCameraLock pattern for viewport changes?
+        //updateProjectionMatrix(width, height);
```
jwoolston: We don't want to update the projection on every frame, we will want to do it in `switchCamera()` and on the size change callback.

jwoolston:
Leave this PR open and make corrections as you develop. We can continue to comment as it evolves and then merge it when its ready.

rajawali/src/main/java/c/org/rajawali3d/engine/Engine.java
```diff
@@ -0,0 +1,131 @@
...
+ * Interface to be implemented for managing the interaction between the rendering context and all
+ * {@link RenderModel}s and {@link RenderView}s associated with it. Rarely, if ever, should user
+ * code implement this interface.
+ *
+ * @author Jared Woolston (Jared.Woolston@gmail.com)
+ */
+public interface Engine {
+
+    /**
+     * Fetches the Open GL ES major version of the EGL surface.
```
rpicolet: OK

rajawali/src/main/java/c/org/rajawali3d/engine/Engine.java
```diff
@@ -0,0 +1,131 @@
...
+     *
+     * @return {@code int} containing the major version number.
+     */
+    int getGLMajorVersion();
```
rpicolet: OK

rajawali/src/main/java/c/org/rajawali3d/engine/Engine.java
```diff
@@ -0,0 +1,131 @@
...
+
+    /**
+     * Fetches the Open GL ES minor version of the EGL surface.
```
rpicolet: OK

rajawali/src/main/java/c/org/rajawali3d/engine/Engine.java
```diff
@@ -0,0 +1,131 @@
...
+     *
+     * @return {@code int} containing the minor version number.
+     */
+    int getGLMinorVersion();
```
rpicolet: OK

rajawali/src/main/java/c/org/rajawali3d/engine/Engine.java
```diff
@@ -0,0 +1,131 @@
...
+     *
+     * @return {@code true} if frames were started/running,
+     *          {@code false} if already stopped (no action taken)
+     */
+    boolean stopFrames();
+
+    /**
+     *
+     */
+    interface SurfaceCallback {
```
rpicolet: OK

rajawali/src/main/java/c/org/rajawali3d/engine/Engine.java
```diff
@@ -0,0 +1,131 @@
...
+        /**
+         * Notifies the client that the render surface dimensions have changed.
+         *
+         * @param width
+         * @param height
+         */
+        void onSurfaceSizeChanged(int width, int height);
+    }
+
+    void addSurfaceCallback(SurfaceCallback callback);
```
rpicolet: Good catch

rajawali/src/main/java/c/org/rajawali3d/engine/Engine.java
```diff
@@ -0,0 +1,131 @@
...
+
+    void removeSurfaceCallback(SurfaceCallback callback);
```
rpicolet: Yep

rajawali/src/main/java/c/org/rajawali3d/engine/Engine.java
```diff
@@ -0,0 +1,131 @@
...
+
+    /**
+     * Retrieves the overall surface height in pixels.
+     *
+     * @return {@code int} The surface height in pixels.
+     */
+    int getSurfaceHeight();
+
+    /**
+     * Registers a {@link RenderView} with the {@link Engine}. The {@link RenderView} will be
```
rpicolet: Good catch

rajawali/src/main/java/c/org/rajawali3d/engine/Engine.java
```diff
@@ -0,0 +1,131 @@
...
+     * notified of the new context and is expected to take any actions it needs to. There is no
+     * guarantee this notification will occur on the GL thread so GL context tasks must be queued
+     * by the {@link RenderView}.
+     *
+     * @param renderModel The {@link RenderModel} object to register.
+     */
+    void addRenderModel(@NonNull RenderModel renderModel);
+
+    /**
+     * Unregisters a {@link RenderView} from the {@link Engine}. The {@link RenderView} will be
```
rpicolet: Yep

rajawali/src/main/java/c/org/rajawali3d/engine/EngineImpl.java
```diff
@@ -35,76 +33,90 @@
  * @author Jared Woolston (Jared.Woolston@gmail.com)
  */
 @NotThreadSafe
-public class RendererImpl implements Renderer, ISurfaceRenderer {
+public class EngineImpl implements Engine, ISurfaceRenderer {
```
rpicolet: I like that idea.  I would think we can abstract out any shared implementations (e.g. frame timing/management?) if/when we add a second render API

jwoolston: Yea. For the time being I dont think it makes sense to worry about an abstract implementation. There is still a fair bit of research on Vulkan to do before we get there.

rajawali/src/main/java/c/org/rajawali3d/engine/EngineImpl.java
```diff
@@ -35,76 +33,90 @@
...
 
     @NonNull
-    protected Context context; // Context the renderer is running in
+    protected Context context; // Android context the engine is running in
 
     protected ISurface surface; // The rendering surface
 
-    protected int defaultViewportWidth;
-    protected int defaultViewportHeight; // The default width and height of the GL viewport
+    protected int surfaceWidth;
```
rpicolet: Could we just make them private and add getters, since they should only be set by the main callback anyway?

rpicolet: Oh duh, yeah, reads could still get messed up, and volatile wouldn't cut it either (even if I had suggested it...).  Will sync!

rajawali/src/main/java/c/org/rajawali3d/engine/EngineImpl.java
```diff
@@ -35,76 +33,90 @@
...
         RajLog.i(context.getString(R.string.renderer_start_header));
         RajLog.i(context.getString(R.string.renderer_start_message));
 
         this.context = context;
         frameRate = getRefreshRate();
 
-        renderables = Collections.synchronizedList(new ArrayList<Renderable>());
+        renderModels = Collections.synchronizedList(new ArrayList<RenderModel>());
+        renderViews = Collections.synchronizedList(new ArrayList<RenderView>());
+
```
rpicolet: Mine too, but I can't tell where it is here...  (I hate github's 8-space tabs. Who uses those?)

Do long lines bug you? I like to keep em at 100 so I never have to scroll horizontally in the IDE. and I changed a few that I was editing anyway, but I don't want to put unwanted fingerprints in the code.

rajawali/src/main/java/c/org/rajawali3d/engine/EngineImpl.java
```diff
@@ -124,13 +136,23 @@ public int getGLMinorVersion() {
     }
 
     @Override
-    public int getDefaultViewportWidth() {
-        return defaultViewportWidth;
+    public void addSurfaceCallback(SurfaceCallback listener) {
```
rpicolet: OK

rajawali/src/main/java/c/org/rajawali3d/engine/EngineImpl.java
```diff
@@ -124,13 +136,23 @@ public int getGLMinorVersion() {
...
+        surfaceCallbacks.add(listener);
     }
 
     @Override
-    public int getDefaultViewportHeight() {
-        return defaultViewportHeight;
+    public void removeSurfaceCallback(SurfaceCallback listener) {
```
rpicolet: OK

rajawali/src/main/java/c/org/rajawali3d/engine/EngineImpl.java
jwoolston: I prefer long lines. All of my screens can easily display 120 characters so thats what i go with.

rajawali/src/main/java/c/org/rajawali3d/engine/EngineImpl.java
```diff
@@ -139,25 +161,27 @@ public boolean isGLThread() {
...
+        renderModel.setEngine(this);
+        renderModels.add(renderModel);
     }
 
     @Override
-    public void addRenderable(@NonNull Renderable renderable) {
-        renderable.setRenderer(this);
-        renderables.add(renderable);
+    public void removeRenderModel(@NonNull RenderModel renderModel) {
+        renderModel.setEngine(this);
```
rpicolet: Um yeah, that would make more sense... OK

rajawali/src/main/java/c/org/rajawali3d/engine/EngineImpl.java
jwoolston: I should add, one other (possibly more performant) option would be to condense these into a holder class that was marked volatile and set by replacing the instance. For example: `volatile Pair<Integer, Integer> dimensions` and then write it by `dimensions = new Pair(width, height);`. I would recommend against using `Pair` and opt for a one off POJO for clarity that can take the primitives vs Boxed values. It should be more performant than acquiring a lock for reads.

rajawali/src/main/java/c/org/rajawali3d/engine/EngineImpl.java
jwoolston: Also, this highlights an argument in favor of using spaces vs tabs. Make sure your IDE is setup to not insert tab characters.

rpicolet:
OK. My Mac screen can handle that easily too. I just tend to use a structure view on the right that cuts down on the available editor window, but I’ll adapt.

rpicolet:
Just checked, not using tabs. Also don’t see any difference in my IDE, both RajLog lines have 8 leading spaces, and I didn’t change those lines anyway?  Not sure what’s up…

rpicolet:
Good idea, will do.

rajawali/src/main/java/c/org/rajawali3d/engine/Engine.java
```diff
@@ -0,0 +1,131 @@
...
+     * @return {@code int} containing the minor version number.
+     */
+    int getGLMinorVersion();
+
+    /**
+     * Checks if the calling thread is the GL thread this {@link Engine} is associated with.
+     *
+     * @return {@code true} if the calling thread is the GL thread.
+     */
+    boolean isGLThread();
```
rpicolet: OK

rajawali/src/main/java/c/org/rajawali3d/engine/EngineImpl.java
rpicolet: Just checked, not using tabs. Also don’t see any difference in my IDE, both RajLog lines have 8 leading spaces, and I didn’t change those lines anyway?  Not sure what’s up…

rpicolet: OK on line length. My Mac screen can handle that easily too. I just tend to use a structure view on the right that cuts down on the available editor window, but I’ll adapt.

rajawali/src/main/java/c/org/rajawali3d/engine/EngineImpl.java
```diff
@@ -220,105 +244,104 @@ public void onRenderSurfaceCreated(EGLConfig config, GL10 gl, int width, int hei
 
     @Override
     public void onRenderSurfaceDestroyed(SurfaceTexture surface) {
-        stopRendering();
+        stopFrames();
     }
 
     @Override
     public void onRenderSurfaceSizeChanged(GL10 gl, int width, int height) {
```
rpicolet: This is from ISurfaceRenderer, which has other GL/EGL dependencies as well...

rajawali/src/main/java/c/org/rajawali3d/engine/EngineImpl.java
```diff
@@ -220,105 +244,104 @@ public void onRenderSurfaceCreated(EGLConfig config, GL10 gl, int width, int hei
...
-        defaultViewportWidth = width;
-        defaultViewportHeight = height;
+        surfaceWidth = width;
+        surfaceHeight = height;
 
-        synchronized (renderables) {
-            for (Renderable renderable : renderables) {
-                renderable.onRenderSurfaceSizeChanged();
+        synchronized (surfaceCallbacks) {
+            for (SurfaceCallback surfaceListener : surfaceCallbacks) {
```
rpicolet: Yes, will do, remind me if I ever forget.

rajawali/src/main/java/c/org/rajawali3d/engine/EngineImpl.java
```diff
@@ -220,105 +244,104 @@ public void onRenderSurfaceCreated(EGLConfig config, GL10 gl, int width, int hei
...
-        }
 
         final long currentTime = System.nanoTime();
-        final long elapsedRenderTime = currentTime - renderStartTime;
-        final double deltaTime = (currentTime - lastRender) / 1e9;
-        lastRender = currentTime;
+        final double deltaTime = (currentTime - lastFrameTime) / 1e9;
+        lastFrameTime = currentTime;
 
-        try {
```
rpicolet: Will do.

rajawali/src/main/java/c/org/rajawali3d/engine/EngineImpl.java
```diff
@@ -220,105 +244,104 @@ public void onRenderSurfaceCreated(EGLConfig config, GL10 gl, int width, int hei
...
-            onRender(elapsedRenderTime, deltaTime);
-        } catch (InterruptedException e) {
-            Log.e(TAG, "Frame skipped due to thread interruption.");
-        }
+        startFrame(deltaTime);
```
rpicolet: I removed it because in general, frames start/stop are independent of the Scene or SceneView lifecycles, and passing this around seemed unnecessary. I did add a getter among the frame control methods, but if there is a use case for passing it around then it makes sense.

rajawali/src/main/java/c/org/rajawali3d/engine/EngineImpl.java
```diff
@@ -220,105 +244,104 @@ public void onRenderSurfaceCreated(EGLConfig config, GL10 gl, int width, int hei
...
             if (fpsUpdateListener != null)
                 fpsUpdateListener.onFPSUpdate(lastMeasuredFPS); // Update the FPS listener
         }
     }
 
-    @Override
-    public void onOffsetsChanged(float xOffset, float yOffset, float xOffsetStep, float yOffsetStep, int xPixelOffset, int yPixelOffset) {
-
-    }
+    protected void startFrame(double deltaTime) {
```
rpicolet: OK on FloatRange, will also incorporate IntRange if we decide to replace elapsed time.

rajawali/src/main/java/c/org/rajawali3d/engine/EngineImpl.java
```diff
@@ -220,105 +244,104 @@ public void onRenderSurfaceCreated(EGLConfig config, GL10 gl, int width, int hei
...
-     * Called by {@link #onRenderFrame(GL10)} to render the next frame. This is called prior to the current scene's
-     * {@link Scene#render(long, double)} method.
-     *
-     * @param ellapsedRealtime {@code long} The total ellapsed rendering time in milliseconds.
-     * @param deltaTime        {@code double} The time passes since the last frame, in seconds.
-     */
-    protected void onRender(final long ellapsedRealtime, final double deltaTime) throws InterruptedException {
-        render(ellapsedRealtime, deltaTime);
+    protected void renderViews() {
+        try {
```
rpicolet: OK

rajawali/src/main/java/c/org/rajawali3d/engine/EngineImpl.java
```diff
@@ -220,105 +244,104 @@ public void onRenderSurfaceCreated(EGLConfig config, GL10 gl, int width, int hei
...
+            for (RenderView renderView : renderViews) {
```
rpicolet: OK

rajawali/src/main/java/c/org/rajawali3d/engine/EngineImpl.java
```diff
@@ -220,105 +244,104 @@ public void onRenderSurfaceCreated(EGLConfig config, GL10 gl, int width, int hei
...
-     * @param ellapsedRealtime {@code long} Render ellapsed time in milliseconds.
-     * @param deltaTime        {@code double} Time passed since last frame, in seconds.
-     */
-    protected void render(final long ellapsedRealtime, final double deltaTime) throws InterruptedException {
-        if (currentRenderable != null) {
-            currentRenderable.render(ellapsedRealtime, deltaTime);
-        } else {
-            Log.w(TAG, "No renderable has been set!");
+    protected void endFrame(double deltaTime) {
+        try {
```
rpicolet: OK

rajawali/src/main/java/c/org/rajawali3d/engine/EngineImpl.java
rpicolet: LOL

rajawali/src/main/java/c/org/rajawali3d/engine/EngineImpl.java
rpicolet: OK, keep em coming.

rajawali/src/main/java/c/org/rajawali3d/engine/EngineImpl.java
jwoolston: We can see how things go with the getter for now.

rajawali/src/main/java/c/org/rajawali3d/engine/EngineImpl.java
```diff
@@ -220,105 +244,104 @@ public void onRenderSurfaceCreated(EGLConfig config, GL10 gl, int width, int hei
...
-    public void onOffsetsChanged(float xOffset, float yOffset, float xOffsetStep, float yOffsetStep, int xPixelOffset, int yPixelOffset) {
-
-    }
+    protected void startFrame(double deltaTime) {
 
-    @Override
-    public void onTouchEvent(MotionEvent event) {
+        // TODO Any engine-level frame tasks, e.g. for adding/removing models or views?
 
+        try {
```
rpicolet: OK

rajawali/src/main/java/c/org/rajawali3d/engine/RenderModel.java
rpicolet: OK

rajawali/src/main/java/c/org/rajawali3d/engine/RenderView.java
rpicolet: This is the kind of thing I had in mind when I suggested a layout manager of some sort, but it seems like it could be feature creep and might turn into a can of worms.  Maybe 2.1 or something, until then, leave it up to the developer?

jwoolston: Yeah I don't think its going to be too much of a can of worms, but definitely enough work to save it for another version. Lets at least make sure the information is available in the meantime in case a developer wants to do something on their own.

rajawali/src/main/java/c/org/rajawali3d/engine/EngineImpl.java
rpicolet: Will do.

rajawali/src/main/java/c/org/rajawali3d/engine/RenderModel.java
rpicolet: Will do as well; Any style/coding conventions/suggestions are welcome

rajawali/src/main/java/c/org/rajawali3d/engine/RenderView.java
rpicolet: OK

rajawali/src/main/java/c/org/rajawali3d/engine/RenderView.java
rpicolet: Can't think of a reason for a null, so @NonNull. Can always setEmpty() to e.g. hide, and we could test for empty before bothering to render

jwoolston: Sounds good.

rajawali/src/main/java/c/org/rajawali3d/engine/RenderView.java
rpicolet: SceneView should never have a null viewport, yeah.

rajawali/src/main/java/c/org/rajawali3d/scene/SceneFrameCallback.java
rpicolet: OK, consistent one way or the other.

rajawali/src/main/java/c/org/rajawali3d/sceneview/SceneView.java
rpicolet: OK

rajawali/src/main/java/c/org/rajawali3d/sceneview/SceneView.java
rpicolet: I'll throw the basic decorator together, then we can add stuff as we think of it

rajawali/src/main/java/c/org/rajawali3d/sceneview/SceneView.java
rpicolet: Makes sense, since otherwise every SceneView.onRenderView() would have to reset the whole state each frame.  Maybe just pass it in and return it both?

jwoolston: yeah, however you want to move it around is fine

rajawali/src/main/java/c/org/rajawali3d/sceneview/SceneView.java
rpicolet: I did this just to see if there is any use case for a view-only display, e.g. some kind of shader-generated geometry or something.  But the spec for that shader could just be a simpler/different kind of RenderModel (e.g. with no scene graph?)  Am I out to lunch?

rajawali/src/main/java/c/org/rajawali3d/sceneview/SceneView.java
rpicolet: Why, you noticed!  How nice! Yeah, default impl...

rajawali/src/main/java/c/org/rajawali3d/sceneview/SceneView.java
jwoolston: In open gl there is no way to create vertices or primitives (triangles, lines, etc) without providing some sort of starting mesh. I would say a simpler/different kind of `RenderModel` makes the most sense in this case.

rajawali/src/main/java/c/org/rajawali3d/sceneview/SceneView.java
rpicolet: I'm just wondering if other elements in the Scene can change during rendering, e.g. materials?  Or is that already handled somehow? Not totally clear on the locking strategy yet

rajawali/src/main/java/c/org/rajawali3d/sceneview/SceneView.java
rpicolet: Yeah, this was just a cut/paste reminder for the TODO; it was commented out originally too

rajawali/src/main/java/c/org/rajawali3d/sceneview/SceneView.java
rpicolet: Will do

rajawali/src/main/java/c/org/rajawali3d/sceneview/SceneView.java
rpicolet: Will do

rajawali/src/main/java/c/org/rajawali3d/sceneview/SceneView.java
jwoolston: It is not, but their safety is intended to be handled separately from the scene (specifically, it comes down to the `ObjectRenderer` applied to an object and the thread safety of whatever holds it

rajawali/src/main/java/c/org/rajawali3d/sceneview/SceneView.java
jwoolston: Dont worry about correcting this yourself. I have not finished the implementation here as what you are doing will affect it so i put it on hold. We will fix later.

rajawali/src/main/java/c/org/rajawali3d/sceneview/SceneView.java
rpicolet: OK, single ctor then, They can pass in an empty Rect if they want

rajawali/src/main/java/c/org/rajawali3d/sceneview/SceneView.java
rpicolet: OK, trial balloon coming.

rajawali/src/main/java/c/org/rajawali3d/sceneview/SceneView.java
rpicolet: OK, sounds good. Scene is getting nice and lean...

rajawali/src/main/java/c/org/rajawali3d/sceneview/SceneView.java
rpicolet: What, me worry?

rajawali/src/main/java/c/org/rajawali3d/sceneview/SceneView.java
jwoolston: Yea thats the goal. Its even leaner in my version compared to the version here you used.

rajawali/src/main/java/c/org/rajawali3d/engine/EngineImpl.java
rpicolet: Preferences->Editor->General->Softwrap gives the best of both worlds...

rpicolet:
First off, sorry for the huge footprint, but I think it's in the nature of the beast.  In this update, I consciously expanded the scope to address several goals, all under the theme of addressing and separating architectural concerns, many of which fell out of the prior decision to separate models and views. These include

 - Names
 - Thread safety and sync (RenderTasks)
 - Frame event processing/delegation
 - Separation of Scenes and their SceneViews
 - Public client APIs versus internally required interfaces
 - Generally separating high-level engine functions from the graphics pipeline implementation (in anticipation of Vulkan)

Rationales and Misc Thoughts/Details

- Names: was using "engine" before, I changed that to "core", as engine is already (better) used to refer to the whole library, and didn't want to confuse users. Also, just rolled "RenderModel" and "RenderView" into Scene and SceneView, as I couldn't identify any other kind of model than a Scene. Also chose RenderControl rather than Renderer to emphasize it's higher-level functionality.  Also, using gles/GLES to avoid conflicts with existing Android openGL/GL/gl names.

- Thread safety and sync: I'm trying to eliminate the need for the different synchronized blocks in the render method, and moving such things as camera and viewport changes into RenderTasks that only need to execute infrequently.  The basic idea is that the actual pipeline render method proper can assume any configuration changes are already in place and can just fly. Minimizes monitor acquisitions for better performance. Not sure what kind of interactions this will have with the SceneModifier approach, may mean certain SceneView RenderTasks will also need to acquire a lock on the Scene/SceneGraph. In general, trying to have a small number architectural sync mechanisms that get used rather than custom solutions for each case. Definitely room for improvement, but its an idea

- Frame event processing: introduced AFrameDelegate to provide propagation of frame start/end events  to Scenes and SceneViews, and to handle their RenderTask queues and animations (seems likely that SceneViews may eventually need animations for viewports and cameras that are not part of a SceneGraph), and their client FrameCallbacks. 

- Scenes vs SceneViews: same basic separation as previously. Seems like Cameras don't necessarily have to be part of a Scene/SceneGraph, unless its a POV camera in the Scene itself or for showing balls/frusta in debug mode, so made Camera a SceneNode but   Put in hooks to manage depth order on viewports, doesn't have to be a 2.0 feature

- Client APIs: Trying to distinguish between public/supported client APIs and internal/unsupported ones, i.e. trying not to mix the two kinds of methods in an interface. Spent some time on the startup control flow, adding RenderControlClient to make sure everything happens in the right order  all the time (e.g. configuring the render context first), and not allowing the client to mess it up, and making it clearer what their responsibilities are when its their turn.  Maybe we also need some annotations for client/supported and/or internal/unsupported methods/interfaces?

- Pipeline independence: seems like now is the right time to put the hinge points in between high-level engine functions and the graphics system. With only cursory looks at Vulkan and my still-limited understanding of GLES, it still seems like there are a small number of potentially well-defined points (context config, loading/allocations, drawing), where the engine can go with whatever pipeline is active. Seems like - if we do decide to support Vulkan - most apps will still want to work on a GLES-only device, so I added RenderContext to have a simple way for clients and the engine to know. I used interfaces to get independence for SurfaceViews/Renderers, some places might be better off using delegation (Geometries and their buffers, Object3Ds and their ObjectRenderers?). Not sure how to get everything constructed (I used factory methods on SceneViews...).

Anyway, all this led to a lot of ripples, so I touched a lot of stuff.  I'm expecting to un-touch some of it when I find out where I've missed the boat.  CamdenHellsBasic is probably the best place to start trying to figure out where the heck I went with all this. 

Enjoy. I need/welcome all feedback you can provide.

rajawali/src/main/java/c/org/rajawali3d/core/ACoreComponent.java
```diff
@@ -0,0 +1,160 @@
...
+        return false;
+    }
+
+    /**
+     *  Runs all {@link RenderTask}s queued for this component
+     */
+    @RenderThread
+    private final void runQueuedTasks() {
+        synchronized (frameTaskQueue) {
+            RenderTask task = frameTaskQueue.poll();
```
ToxicBakery: instead of null checks (I hate null when it can be avoided) what about

```java
while (!frameTaskQueue.isEmpty()) {
  RenderTask task = frameTaskQueue.removeFirst();
  task.run();
}
```

The implementation of `isEmpty()` uses the internal size counter of a `LinkedList` and checks it against zero. Effectively this is what `poll()` does internally so instead of using `poll()` which would now cause a redundant size check (which size == 0 causes null to be returned) you can just use `removeFirst()`.

rajawali/src/main/java/c/org/rajawali3d/core/AFrameDelegate.java
```diff
@@ -0,0 +1,234 @@
...
+    }
+
+    @Override
+    public final boolean isEnabled() {
+        return isEnabled;
+    }
+
+    @Override
+    public boolean addFrameCallback(@NonNull final FrameCallback callback) {
+        final RenderTask task = new RenderTask() {
```
ToxicBakery: Internally I would like to see concrete implementations vs anonymous implementations whenever possible. In the examples I don't care too much but in the internal engine I think it would benefit us to be as concise as possible and anonymous classes add a lot of noise.

rajawali/src/main/java/c/org/rajawali3d/core/AFrameDelegate.java
```diff
@@ -0,0 +1,234 @@
...
+
+    //
+    // ACoreComponent method overrides
+    //
+
+    @RenderThread
+    @Override
+    @CallSuper
+    public void onFrameStart(double deltaTime) throws InterruptedException {
+        if (!isEnabled) {
```
ToxicBakery: Is there any way that this can happen either somewhere else or after the super call. For sanity of debugging I'm not a fan of placing anything before super and I would be surprised if PMD doesn't complain about this for the same reason.

rajawali/src/main/java/c/org/rajawali3d/core/ARenderControl.java
```diff
@@ -0,0 +1,422 @@
...
+        // Propagate to SceneViews...//
+        for (int i = 0, j = sceneViews.size(); i < j; i++) {
+            ((ASceneView)sceneViews.get(i)).onRender();
+        }
+    }
+
+    @RenderThread
+    private void updateFPS() {
+        ++frameCount;
+        if (frameCount % 50 == 0) {
```
ToxicBakery: What is the reason for this being 50 vs any other number?

rajawali/src/main/java/c/org/rajawali3d/core/AFrameDelegate.java
jwoolston: This is probably my fault. Over the years I have gotten lazy with stuff like this and started throwing in anonymous classes for one or two line runnables in spite of how much I hate them. Really, this is an ideal candidate for a Lambda, but we cant use Java 8.

rajawali/src/main/java/c/org/rajawali3d/core/ARenderControl.java
jwoolston: Well, specifically if the goal is to update once per second, it should be the operating frame rate.

rajawali/src/main/java/c/org/rajawali3d/annotations/RenderThread.java
```diff
@@ -7,12 +7,12 @@
...
- * Method level annotation to indicate that a method is expected to run on the GL thread.
+ * Method level annotation to indicate that a method is expected to run on the render thread.
  *
  * @author Jared Woolston (Jared.Woolston@gmail.com)
  */
 @Documented
 @Retention(RetentionPolicy.CLASS)
 @Target({ElementType.METHOD, ElementType.FIELD})
-public @interface GLThread {
+public @interface RenderThread {
```
jwoolston: Good catch

rajawali/src/main/java/c/org/rajawali3d/core/ACoreComponent.java
```diff
@@ -0,0 +1,160 @@
...
+import java.util.LinkedList;
+import java.util.Queue;
+
+/**
+ * Base class for all core components; enables common render thread tasking
+ *
+ * @author Randy Picolet
+ */
+
+abstract class ACoreComponent implements CoreComponent {
```
jwoolston: I am generally not a fan of the `A` or `I` prefixes, in spite of the fact I am guilty of having used them before. This basically goes along with my desire to remove Hungarian Notation from the library.

rajawali/src/main/java/c/org/rajawali3d/core/AFrameDelegate.java
ToxicBakery: We could use Kotlin though 😁

jwoolston: Frankly, even though there are many features of Kotlin I wish Java had, there are many I wish programming languages didnt and I am still on the fence about how much of my time I wish to invest in it.

rajawali/src/main/java/c/org/rajawali3d/core/ACoreComponent.java
rpicolet: Will do.

rajawali/src/main/java/c/org/rajawali3d/core/AFrameDelegate.java
rpicolet: So for now, I should just add an inner class extending RenderTask for each task type? And adopt avoiding anonymous classes as a convention in similar situations?

rajawali/src/main/java/c/org/rajawali3d/core/AFrameDelegate.java
rpicolet: Yeah, I could move the enable check to the `RenderControl` delegation loops before the call is made, like this method does for animations, or I could just leave this method abstract in `ACoreComponent` and call `runQueuedTasks()` directly from here and in RenderControl, or introduce a `doCoreComponentStuffForFrameStart()` method so if the logic changes in the future there is still only one place to change. 

Out of curiosity, what kinds of problems can come up by doing something before calling super methods?  You mentioned debugging, but I don't see the issues/

And you made me look up PMD. 👓  Seems like a super-lint.

rajawali/src/main/java/c/org/rajawali3d/core/ARenderControl.java
rpicolet: Easy enough to use the frame rate. Will do.

rajawali/src/main/java/c/org/rajawali3d/core/AFrameDelegate.java
ToxicBakery: The issue comes from, as a developer, you expect stacktraces of classes that extend to call super first so without expecting every single implementation life cycle events become difficult to understand when they don't happen in the natural order of `let the parent do the work before I do my work`

I believe in most situations like this @jwoolston and I would typically agree on is try to use a delegate instead of extending a method to hack in life cycle changes. I'm not sure how or if that can apply here.

rpicolet:
@jwoolston Yeah, I hope my intentions are mostly obvious from looking at the code, even where there are incomplete thoughts and stupid coding errors or oversights. When you find places where the rationale is opaque/confusing, lemme know.

rajawali/src/main/java/c/org/rajawali3d/annotations/RenderThread.java
rpicolet: Yeah, I'm assuming either Android or we will provide an equivalent to GLSurfaceView and its GL Thread stuff, even if the pipeline itself uses multiple workers

rajawali/src/main/java/c/org/rajawali3d/core/ACoreComponent.java
rpicolet: I don't particularly care what naming convention is used, but what should we use when there is both an interface and an abstract shared implementation of that interface, as there is here?

rajawali/src/main/java/c/org/rajawali3d/core/AFrameDelegate.java
rpicolet: I guess it's like most such issues, a trade-off between concerns.  I see that it would be easier to understand stack traces and control flows if you can assume super-first, but avoiding overriding of super methods whenever that convention might be violated seems limiting too. I don't really get the "natural order" argument, except for constructors. Don't C++ destructors always run super-last as the "natural order"? In between, it seems arbitrary to insist on super first.  

For example, in my app's Android activity life cycle, I need to deal with model initialization before fragments get their onStarts delegated by the Android parent class, and to disable model-observer notifications before allowing fragments to get onPause events. And in composite-patterns, it seem awkward to hold child container classes (which might form a subtree themselves) responsible to delegate the nested child iteration; instead, it gets confusing about which node is responsible for what, and/or the logic gets duplicated. 

In this case, I think I'd prefer to just check for enabled in the caller. That avoids your concern here and we can debate the rest of it over a beer some day. Wait, no, that could ruin a really good quaft...

jwoolston: In this case I am with @rpicolet, I routinely add code prior to calling a super method for the very specific reason of manipulating the data the super method will be acting on, and in my interpretation of OOP, this is very much an intended consequence of inheritance.

rajawali/src/main/java/c/org/rajawali3d/core/ACoreComponent.java
jwoolston: In this case I think `BaseCoreComponent` might be an appropriate name.

jwoolston:
Will do, just give me some time to take a proper pass at really understanding what you were doing so that I can post more intelligent questions/comments on it.

rpicolet:
What, it might take you a couple days? It never takes me any time at all. I'm always fast, fast, fast, er, um... yeah.

No hurry, obviously.  I'm going to start looking at the render pass stuff again. Remember render passes?

rajawali/src/main/java/c/org/rajawali3d/core/ACoreComponent.java
rpicolet: <tl;dr>
On Hungarian notation and naming conventions

I never liked encoding data types in variable names; it is almost always unreadable, cognitively distracting, a maintenance headache, and rarely useful as fundamental types (number vs string vs boolean vs reference) can usually be inferred and Java is type safe anyway. I've rarely seen it in Java code,

I've never seen data return types encoded in Java method names (maybe in JNI? I forget).

On the other hand, role/purpose indicators go to the semantics of an identifier, and are especially useful in reducing name collisions and lengths.

While I can live without the m or s prefix on instance/class member field names, I prefer to use them, just as I prefer to use all caps/underscores for static finals. I don't really like using `this.var = var;` all the time in constructors and setters (it is too easy to forget to use "this" later on in the method), and knowing at a glance the role of a variable - whether it's a local or parameter versus an instance member or static/class member - is simply more critical to me than its declaration details (data type, visibility, and other modifiers); it helps me know _where_ I need to go to find those when I care. I understand that IDEs visually provide role information via color and font, and full declaration info on hovers etc, but when looking at code outside of an IDE like here in github, it actually requires you to manually search through files for it (especially troublesome are inherited fields, if you don't even know it's a field) and without indicators in the name you don't get a clue about where to start.  And even in a IDE there is no standard anyway; one might use blue, another green. one bold, one italic, one underscores for different things, not to mention dev customizations.

With type names, you have the basic semantic concept (the name itself, a critical choice to me), and multiple aspects of that concept's definition and refinement in both interface and class hierarchies to deal with. Removing all shorthand indicators of interface vs abstract implementation vs concrete implementation doesn't seem like an improvement to me.  It's already hard enough to provide clear distinctions consistently and efficiently. Name collisions will increase without some kind of disambiguation for these different roles.  A and I seem simple, and in particular A reads well in English. I suppose we could use full words - AbstractFoo or BaseFoo or SomeFoo, but why? I and A are short, clear, and unambiguous de facto standards. They are not data types, but core aspects of the identification of the named artifacts that distinguish one from another.

If we don't use those, we should come up with some functionally equivalent conventions.  

</tl:dr>

PR #1842, opened 2017-02-16
Fix set surface ilegal state exception

SandroMachado:
I am using this library and I am getting multiple times the following crash:

```
Fatal Exception: java.lang.IllegalStateException
       at android.media.MediaPlayer._setVideoSurface(MediaPlayer.java)
       at android.media.MediaPlayer.setSurface(MediaPlayer.java:760)
       at org.rajawali3d.materials.textures.StreamingTexture.add(StreamingTexture.java:83)
       at org.rajawali3d.materials.textures.TextureManager.taskAdd(TextureManager.java:112)
       at org.rajawali3d.materials.textures.TextureManager.taskReload(TextureManager.java:211)
       at org.rajawali3d.renderer.Renderer.initScene(Renderer.java:366)
       at com.sandro.aerial.RendererActivity.onRenderSurfaceSizeChanged(RendererActivity.java:206)
       at org.rajawali3d.view.SurfaceView$RendererDelegate.onSurfaceChanged(SurfaceView.java:232)
       at android.opengl.GLSurfaceView$GLThread.guardedRun(GLSurfaceView.java:1564)
       at android.opengl.GLSurfaceView$GLThread.run(GLSurfaceView.java:1267)
```

This PR should fix the issue and it should be the appropriated solution since our callback that calls setSurface gets hit multiple times throughout the setup of playback, by the time start is called, the surface has been set again without throwing the exception.

### DIFF
--- a/rajawali/src/main/java/org/rajawali3d/materials/textures/StreamingTexture.java
+++ b/rajawali/src/main/java/org/rajawali3d/materials/textures/StreamingTexture.java
@@ -19,6 +19,8 @@ import android.opengl.GLES11Ext;
 import android.opengl.GLES20;
 import android.view.Surface;
 
+import org.rajawali3d.util.RajLog;
+
 import java.io.IOException;
 
 public class StreamingTexture extends ATexture {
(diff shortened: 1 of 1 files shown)


SandroMachado:
@jwoolston can you please take a look at this PR. I am having a lot of issues with this crash log.

Thanks :)

rpicolet:
Your proposed solution (swallow/logging the exception) is not appropriate, as there could be multiple causes for the exception, and simply failing to set the new surface on the MediaPlayer is pointless as it will at least look broken even if it somehow keeps running.

I might be wrong, but I would guess you could do something in com.sandro.aerial.RendererActivity.onRenderSurfaceSizeChanged() to check/adjust the MediaPlayer into an acceptable state prior to Rajawali's reloading/adding the texture...

jwoolston:
Well said @rpicolet

SandroMachado:
I am probably missing something, but AFAIK I need to call the `super` every time the callback get `onRenderSurfaceSizeChanged`, and not only when the `MediaPlayer` object is ready. Or I am missing something?

jwoolston:
Our general philosophy with Rajawali is to propagate all exceptions back to the user. This is common in libraries unless the library itself is going to attempt to fix it, which we arent. If all you want to do is bury the exception, you should be able to accomplish this in your own code.

Issue #1850, opened 2017-02-22
Tilted camera images on the ScreenQuad


entkler:
#### Rajawali Version or Branch
1.1.777 (even later builds like 1.1.863)

#### Device and Android Version
Project Tango tablet (Android 4.4)

#### Summary
I tried to stream camera images on a "ScreenQuad" Object this worked but just that the images are tilted upside down. This happend using Rajawali 1.1.777 (or even the later master builds like 1.1.863). Any idea how i can fix this ? thanks for the help

contriteobserver:
Submitted path to adress UV coordinate issue.

jwoolston:
Resolved via #1854

Issue #1852, opened 2017-02-27
move object with Arcball Camera


thelordvampire:
Rajawali Version 1.1.777

how to pick and move object with arcball camera? if i use normal camera, i can easily pick and move an object but i can't  take camera flying around with z axis(world coordinate system, not camera coordinate system) and always see the point(0,0,0). The solution arcball camera is so great,it is simple to do that, but i can't pick and move an object? What should i do now? Thanks for your help

thelordvampire:
I found enableLookAt() in Camera so i could take camera always see one point. But  why i can't pick and move object with arcball camera?

jwoolston:
Arcball camera does not enforce usage of much of the base API. This will be resolved in the future.

PR #1854, opened 2017-03-06
UV orientation in ScreenQuad primitive now matches that of Plane.

contriteobserver:
fixes Tilted camera images on the ScreenQuad #1850

### DIFF
--- a/rajawali/src/main/java/org/rajawali3d/primitives/ScreenQuad.java
+++ b/rajawali/src/main/java/org/rajawali3d/primitives/ScreenQuad.java
@@ -184,10 +184,10 @@ public class ScreenQuad extends Object3D {
 				vertices[vertexCount + 2] = 0;
 
 				if (mCreateTextureCoords) {
-					float u = (float) mSegmentsW - ((float) i / (float) mSegmentsW);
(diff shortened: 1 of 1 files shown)


jwoolston:
Thanks @contriteobserver. I have dealt with this myself but never got around to fixing it.

rpicolet:
I think I was the last one to touch those coordinates. Prior to my  change, u used `1.0 - u` and v used `1.0 - v`, but while testing the combination of color picking with postprocessing effects, I changed the v coordinate (to just `v`, not `1.0 - v`) to get everything working.  So if this is now the correct mapping, there may be some other places where things need to also change, maybe the color picking glReadpixels call (it uses the flipped version for y, i.e. `viewport.height - y`), and/or somewhere in postprocessing.  But I really don't understand all the coordinate mapping choices here and throughout the library yet, and how they are supposed to work together.

jwoolston:
> But I really don't understand all the coordinate mapping choices here and throughout the library yet, and how they are supposed to work together.

They are a mess and I don't want to deal with them unless someone pays me to. Our 2.0 work should take care of it.

jwoolston:
#1852

PR #1877, opened 2017-04-19
added ScreenMappingRatio setter and getter to ArcballCamera

contriteobserver:
responding to #1868 

added `ScreenMappingRatio` setter and getter to `ArcballCamera`, value is 1,
setting this ratio to -1 accomodates the use case of looking at all sides small object
from the outside, without having an environment to give context to the camera movement.
Values from -1 to 1 are accepted, setting to 0 locks the `ArcballCamera` at the current orientation.

### DIFF
--- a/rajawali/src/main/java/org/rajawali3d/cameras/ArcballCamera.java
+++ b/rajawali/src/main/java/org/rajawali3d/cameras/ArcballCamera.java
@@ -7,6 +7,8 @@ import android.view.MotionEvent;
 import android.view.ScaleGestureDetector;
 import android.view.View;
 
+import android.support.annotation.FloatRange;
+
 import org.rajawali3d.Object3D;
 import org.rajawali3d.math.MathUtil;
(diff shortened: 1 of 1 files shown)


rajawali/src/main/java/org/rajawali3d/cameras/ArcballCamera.java
```diff
@@ -73,6 +74,16 @@ public void setProjectionMatrix(int width, int height) {
         super.setProjectionMatrix(width, height);
     }
 
+    public void setScreenMappingRatio(float ratio) {
```
ToxicBakery: Should have `@FloatRange(from=-1.0, to=1.0)`

contriteobserver: committed FloatRange change

contriteobserver:
related to #2014

Issue #1885, opened 2017-05-02
Updating custom shaders - black screen on launch


mukulabdagiri:
#### Rajawali Version or Branch
Master

#### Device and Android Version
Samsung Galaxy 6, Android version 6.0.1

#### Summary
I recently upgraded my project to use latest master version of Rajawali library. It was previously using old 0.9 tag. I updated all my code to use latest library features. But now the project just shows black screen upon launch. So i think the shaders are not set up correctly as I did not update them. Below is the shader code which was working fine with the 0.9 tag. Kindly let me know if the below code will work with the current Rajawali version or if I need to modify it.

#### Steps to Reproduce

```
//BackgroundImage class function to initialize image

public void init(AppRenderer renderer, Context context, int resId)
	{
		try{
	    	FadeMaterial planeMat = new FadeMaterial();
			Bitmap bitmap = BitmapFactory.decodeResource(context.getResources(), resId);
			Texture mutableTexture = new Texture("mutable", bitmap);
			planeMat.addTexture(mutableTexture);
			plane = new Plane(100, 56, 1, 1);
			plane.setScale(PLANE_SCALEFACTOR);
			plane.setMaterial(planeMat);
			plane.setTransparent(true);
			renderer.getCurrentScene().addChild(plane);
		}catch(TextureException e){
			e.printStackTrace();
		}
		
		timer = 0;
		isActive = false;
	}

//FadeMaterial class to fade images

package com.mycompany.myapp;

import org.rajawali3d.materials.Material;
import org.rajawali3d.materials.shaders.FragmentShader;
import org.rajawali3d.materials.shaders.VertexShader;

//previously using - public final class FadeMaterial extends SimpleMaterial
public final class FadeMaterial extends Material {
	
	protected static final String mVShader = 
		    "uniform mat4 uMVPMatrix;\n" +

		    "attribute vec4 aPosition;\n" +
		    "attribute vec2 aTextureCoord;\n" +
		    "attribute vec4 aColor;\n" +

		    "varying vec2 vTextureCoord;\n" +
		    "varying vec4 vColor;\n" +      

		    "void main() {\n" +
		    "   gl_Position = uMVPMatrix * aPosition;\n" +
		    "   vTextureCoord = aTextureCoord;\n" +
		    "   vColor = aColor;\n" +
		    "}\n";

		protected static final String mFShader = 
		    "precision mediump float;\n" +

		    "varying vec2 vTextureCoord;\n" +
[...]

PR #1887, opened 2017-05-06
fix for creating custom shaders issue #1885

contriteobserver:
`Material.createShaders()` is nolonger overwriting custom shader strings


ToxicBakery:
Not sure I agree with this, it would be wiser to have an a function that allows optimization of callers vs changing a builder to only sometimes build.

contriteobserver:
tried another approach, see #1896

jwoolston:
Declining in favor of #1896

Issue #1858, opened 2017-03-10
RVect4 component w


TBog:
#### Rajawali Version or Branch
master

#### Device and Android Version
device SM-T705 android 5.0.2

#### Summary
RVect4 **w** component should return the same type of ShaderVar as **a** or **z** component

#### Steps to Reproduce
GL_POSITION.w().assign( GL_POSITION.z() );

#### Trace or Log Output
generated shader code for the above will look like this:
vec4 gl_Position.w = gl_Position.z;

and produce error:
[org.rajawali3d.materials.Material] Could not compile vertex shader:
Shader log: 0:29: L0006: Illegal identifier name 'gl_Position'
0:29: L0001: Expected token ';', found '.'

jwoolston:
The shaderlog is telling you you cant define a new variable with the name gl_Position. This is because it is a reserved name in GLSL.

TBog:
I am not defining, I am setting a value.
If I write `GL_POSITION.a().assign( GL_POSITION.z() );` it works as expected!

jwoolston:
`vec4 gl_Position.w = gl_Position.z;` The fact that you have the data type at the start of the line shows you are defining.

jwoolston:
Sorry, was early. The shader builder is generating the wrong GLSL and trying to declare `gl_Position` in that case rather than simple assignment (why the `vec4` shows up).

contriteobserver:
using this test case to make the issue manifest
```
    private class ConstructedShaderRenderer extends Renderer {
        Object3D obj;

        public ConstructedShaderRenderer(Context context) {
            super(context);
            RajLog.setDebugEnabled(true);
        }

        @Override
        protected void initScene() {
            try {
                getCurrentScene().setBackgroundColor(Color.BLUE);

                VertexShader vertexShader = new VertexShader() {
                    private RMat4 uMVPMatrix;
                    private RVec4 aPosition;

                    @Override
                    public void initialize() {
                        super.initialize();

                        addPrecisionQualifier(DataType.FLOAT, Precision.MEDIUMP);
                        // -- uniforms
                        uMVPMatrix = (RMat4) addUniform(DefaultShaderVar.U_MVP_MATRIX);
                        // -- attributes
                        aPosition = (RVec4) addAttribute(DefaultShaderVar.A_POSITION);
                    }

                    @Override
                    public void main() {
                        RVec4 location = new RVec4("location");
                        location.assign(uMVPMatrix.multiply(aPosition));
[...]

PR #1888, opened 2017-05-09
added xyzw() and rgba() functions to RVect4, and fixed w() assignment.

contriteobserver:
This addresses the issue raised in #1858

None too pleased with the `xyzw` nomenclature, but it seems the way forward, as `RVect4` inherits from `RVect3`

### DIFF
--- a/rajawali/src/main/java/org/rajawali3d/materials/shaders/AShaderBase.java
+++ b/rajawali/src/main/java/org/rajawali3d/materials/shaders/AShaderBase.java
@@ -507,11 +507,28 @@ public abstract class AShaderBase {
 			super(name, dataType, value);
 		}
 		
-		public ShaderVar w()
+		public ShaderVar wxyz()
 		{
 			ShaderVar v = getReturnTypeForOperation(mDataType, mDataType);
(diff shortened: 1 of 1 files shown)


rajawali/src/main/java/org/rajawali3d/materials/shaders/AShaderBase.java
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright 2013 Dennis Ippel
```
ToxicBakery: why did the copyright change?

contriteobserver:
It was a typo, I thought I had fixed it in the second submission, but I forgot an asterisk

rajawali/src/main/java/org/rajawali3d/materials/shaders/AShaderBase.java
contriteobserver: was going to add  comment and then changed my mind, the copyright line got changed by mistake, commited a change to fix the typo.

jwoolston:
@contriteobserver Thankyou for your continued efforts to help us not only with responding in issues but also submitting fixes.

Issue #1634, opened 2016-03-21
Object picking for augmented reality.


conect:
Hi I've got a Renderer which inherits from https://github.com/googlesamples/tango-examples-java/blob/8d315d38007265e45a4089379a3fc7e6da4ee18e/java_examples_utils/app/src/main/java/com/projecttango/rajawali/ar/TangoRajawaliRenderer.java . Now I want to implement object picking in the augmented reality view. Problem is that the object picking works fine, as long as the backgroundsquad is not visible (mBackgroundQuad.setVisible(false)) but as soon as the squad is rendered, objectcolorpicking doesn't work anymore. Could someone tell me why?

Edit: To reproduce do following:
1. Clone https://github.com/conect/tango-examples-java.git
2. Checkout objectpickingsample
3. Run the augmented reality example.

If you click on the green cube there should be a debugmessage, but there isnt.

Thanks a lot.

jwoolston:
As I mentioned in the other ticket, you need to be more descriptive with your issues. I have no idea what other stuff you have added to the scene but a screen quad should be thought of as holding up a drawing in front of the camera. The renderer they added there is simply streaming a texture to rajawali from the tango subsystem, not providing geometry to rajawali - or at least not to that renderer.

conect:
Yes the image frame is supposed to be merely visible in the background. I've placed two cubes and a plane in the scene. And registered them for object pickin this works fine as long as the screen quad is not visible.

jwoolston:
You mentioned using ray picking. Are you doing this yourself or with the `RayPicker` class? The ray picking in rajawali is incomplete and hasn't been touched along with the other upgrades so I'm suprised if it works under any conditions. Have you tried using color picking? I say that but the color picking in the Rajawali version they used in their examples is broken as well and was only recently fixed.

conect:
Sry i use object colorpicking. Thouhght it was the same principle...

jwoolston:
Color picking and ray picking are different algorithms. The color picking in that version of the library is known to be broken. I will be sending them a PR for updating to the latest version of rajawali sometime this week but I do not know if/when they will accept it. In the meantime, you could always force the update yourself.

conect:
Yes of course i'm already using the newest version now the colorpicking works as long as the screenquad is not visible.

jwoolston:
I will have to try this out and see if i can reproduce the problem then to figure out whats going on.

conect:
Thx a lot!

[...]

PR #1895, opened 2017-05-24
Move depth test for color picking to per-object

adamantivm:
Ref: #1634 

This change allows color picking to work when there is a ScreenQuad involved in the scene, by properly rendering said quad in the background, like it happens with the generated scene.

In order to test this, please use branch `color-picker` in [my fork of the Tango Java samples](https://github.com/adamantivm/tango-examples-java/tree/color-picker). This branch includes modification to its `build.gradle` so that it will use a latest Rajawali 1.1 snapshot, so that it is possible to build the Rajawali fix and test it against this sample.

Please clone my fork, build the app `java_augmented_reality_example` and run it on a Tango-enabled device. You should see a rotating earth with a revolving moon against the backdrop of the live camera image. If you click on the earth, you should see the following in the logcat:

```
05-24 12:32:26.336 10329-10329/com.projecttango.experiments.augmentedrealitysample D/AugmentedRealityRenderer: ---> pick attempt at (1474.0,420.0
05-24 12:32:26.394 10329-10360/com.projecttango.experiments.augmentedrealitysample D/AugmentedRealityRenderer: Picked: org.rajawali3d.primitives.ScreenQuad@648f81d8
```

Before the Rajawali fix, you will always see the `ScreenQuad` being picked. After the fix, the proper sphere for the moon or earth should be picked.

@jwoolston I hope this makes it simple to verify. Please do let me know if anything here is problematic and I'll be happy to work towards making it easier. Thanks again for your support.

### DIFF
--- a/rajawali/src/main/java/org/rajawali3d/Object3D.java
+++ b/rajawali/src/main/java/org/rajawali3d/Object3D.java
@@ -426,7 +426,17 @@ public class Object3D extends ATransformable3D implements Comparable<Object3D>, 
 				}
 			}
 
-			// Blending and depth testing are set up globally in Scene.doColorPicking()
+			// Blending test is set up globally in Scene.doColorPicking()
+
+			// Depth testing is set-up per-object in order to avoid ScreenQuads to overshadow other
+			// objects, see https://github.com/Rajawali/Rajawali/issues/1634
+			if (!mEnableDepthTest) GLES20.glDisable(GLES20.GL_DEPTH_TEST);
+			else {
+				GLES20.glEnable(GLES20.GL_DEPTH_TEST);
+				GLES20.glDepthFunc(GLES20.GL_LESS);
+			}
+
+			GLES20.glDepthMask(mEnableDepthMask);
 
 			// Material setup is independent of batching, and has no need for
 			// shader params, textures, normals, vertex colors, or current object...
--- a/rajawali/src/main/java/org/rajawali3d/scene/Scene.java
+++ b/rajawali/src/main/java/org/rajawali3d/scene/Scene.java
@@ -1172,12 +1172,6 @@ public class Scene {
 		// Set background color (to Object3D.UNPICKABLE to prevent any conflicts)
 		GLES20.glClearColor(1.0f, 1.0f, 1.0f, 1.0f);
 
-		// Configure depth testing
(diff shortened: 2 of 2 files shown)


rpicolet:
If I understand how this works correctly, perhaps it should be mentioned/documented that this solution also depends on the order in which objects are added to the scene, i.e. it relies on the screen quad (or other non-depth-tested objects) being added (and so rendered) first.

adamantivm:
@rpicolet if the screen quad is not added at the back of the scene, it will overlay everything in the rendering as well, so I think the behavior will be as expected and evident, no?

rpicolet:
@adamantivm Sure, probably, but what's obvious to one may be confusing to another.  It's an undocumented assumption, like dozens of others, and I wasn't proposing any code change for that.  This thread is already more explanation than many such assumptions get, and I was mostly thinking about "artistic" use of non-depth-tested objects that were mentioned previously.

adamantivm:
@rpicolet I appreciate the comment. I personally don't see how it could possibly produce a confusing result but I am happy to add any additional documentation (where?) that people agree would be useful.

rpicolet:
Well, it was confusing for me at first.  And as I was the one who wrote the global depth test code, there ya go; we all have different experience bases, and I erroneously and unconsciously assumed VR apps (which I have no experience with) were using a transparent background in the surface view over a regular camera view rather than a screen quad.  Anyway, as I said, I think this discussion is enough docs for now.

I'm more concerned about the deletion of the glClearDepth() call, that could have unintended consequences if some other use case (don't ask me what that might be) sets it to something else elsewhere in the code.

And thanks for your contributions to Rajawali and my education!

jwoolston:
@adamantivm I am starting to take a look at this now. I will let you know if I have any issues. I probably need to wait for my tango device to charge.

jwoolston:
Everything seems to check out. @adamantivm as always, thank you for constructive and useful efforts and most importantly, making it not take undo work on our end to observe the problem and fix. Your help is always appreciated.

adamantivm:
Thanks to you @jwoolston for all your hard work bringing this library out to the community. My part is simply a small part of what anyone should do. Your receptiveness as a maintainer also makes an important difference. Please, keep it up.

PR #1896, opened 2017-05-25
another approach to not building a shaderstring when a shaderstring is provided

contriteobserver:
an alternative fix for issue #1885


jwoolston:
I very much prefer this fix to the other one, and this is in fact how I manage this myself. Thankyou again.

Issue #1897, opened 2017-05-26
Unexpected Quaternion Division Behaviour


contriteobserver:
If one adds the following test to our testSuite,
```
    @Test
    public void testDivision() throws Exception {
        final Quaternion q = new Quaternion(4d, 3d, 2d, 1d);
        final Quaternion r = new Quaternion(1d, 2d, 3d, 4d);
        final Quaternion p = q.clone().multiply(r.clone());

        // from p = q * r follows that q = p * (r ^ -1) and r = (q ^ -1) * p
        assertEquals(
            q.toString(),
            p.clone().multiply(r.clone().pow(-1)).toString()
        );
        assertEquals(
            r.toString(),
            q.clone().pow(-1).multiply(p).toString()
        );
    }
```
it unexpectedly fails with:
```
org.rajawali3d.math.QuaternionTest > testDivision FAILED
    org.junit.ComparisonFailure: expected:<Quaternion <w, x, y, z>: <[4.0, 3.0, 2.0, 1.0]>> but was:<Quaternion <w, x, y, z>: <[21.908902300206645, 16.431676725154986, 10.954451150103324, 5.477225575051666]>>
```
Granted, the strings may not match exactly, but one would expect that the results should be close.

jwoolston:
This is unexpected because quaternion power functions are not like they are on scalars. See [https://math.stackexchange.com/a/939288/80969](https://math.stackexchange.com/a/939288/80969) for details. Essentially: `q.pow(-1) != 1/q`.

jwoolston:
@contriteobserver Please shoot me an email if you could.

contriteobserver:
It turns out that  `q.pow(-1) = 1/q`, but our implementation of pow() is only valid for unit Quaternions.

PR #1898, opened 2017-05-26
additional Quaternion unit tests

contriteobserver:
added Quaternion commutativity, associativity, and distributivity unit tests
added a division unit test commented out because it (unexpectedly) fails, see #1897


jwoolston:
As a general rule, unit tests should not compare mathematical results by examining the result of `toString()`.  `toString()` has no contract specifying what it may or may not produce, and as such there is no guarantee that it will properly reflect the state of the object(s) under test.

jwoolston:
Please also see my comment in #1897 regarding the division test.

jwoolston:
Please remove the use of `toString()` for checking the quaternion state and use the internal state directly. See above for an explanation why.

Issue #1899, opened 2017-05-27
Quaternion.pow works only for unit Quaternions


contriteobserver:
#### Rajawali Version or Branch
master

#### Summary
given a quaternion p(1,2,3,4), the result of p.pow(2) should be <-28,4,6,8>,
same as the result of p.clone().multiply(p)

#### Steps to Reproduce
Add the following test case to the unit tests:
```
    public void testPow2() throws Exception {
        final Quaternion q = new Quaternion(1d, 2d, 3d, 4d);
        final String s = q.clone().pow(2).toString();
        assertEquals("Quaternion <w, x, y, z>: <-28.0, 4.0, 6.0, 8.0>", s);
    }
```

PR #1900, opened 2017-05-27
Quaternion.pow now works for non-unit quaternions

contriteobserver:
fix for Quaternion.pow works only for unit Quaternions #1899


jwoolston:
@contriteobserver can you fix the conflicts here?

contriteobserver:
Sorry, I accidentally merged it. I have never used this interface before.

jwoolston:
Thats fine. I would have merged it once the conflicts were resolved anyway.

Issue #1904, opened 2017-05-31
Skybox size is hard coded


alex199568:
#### Rajawali Version or Branch
Apparently it is hard coded in all versions and is equal to 700

#### Summary
As my camera moves too far from the objects in the scene (when the distance between the camera and objects is greater than 350, because hard coded value is 700) the objects get obscured by too small skybox, which I cannot change, because the size is hard coded

```
public void setSkybox(int posx, int negx, int posy, int negy, int posz, int negz) throws TextureException {
		synchronized (mCameras) {
			for (int i = 0, j = mCameras.size(); i < j; ++i)
				mCameras.get(i).setFarPlane(1000); // why 1000? Why setSkybox changes far plane of all cameras?
		}
		synchronized (mNextSkyboxLock) {
			mNextSkybox = new Cube(700, true); // Why 700? What if I need it to be 800? or 600?
			int[] resourceIds = new int[] { posx, negx, posy, negy, posz, negz };
			
			mSkyboxTexture = new CubeMapTexture("skybox", resourceIds);
			((CubeMapTexture)mSkyboxTexture).isSkyTexture(true);
			Material mat = new Material();
			mat.setColorInfluence(0);
			mat.addTexture(mSkyboxTexture);
			mNextSkybox.setMaterial(mat);
		}
	}
```

contriteobserver:
submitted a patch to set skybox geometry

rami-dabain:
causing problems on large worlds, can view more than 350 units, need to view up to 18000

TBog:
The skybox should be a box of size 1, be drawn last so that we don't redraw pixels, centred on camera position. I can provide a patch with this but it involves material plugin and I'm not sure if it's user friendly enough

jwoolston:
Closing as resolved/stale. Please inform if issues persist.

PR #1909, opened 2017-06-09
added ability to set skybox geometry

contriteobserver:
addresses issues raised in #1904
also partially cleans up skybox implementation

### DIFF
--- a/rajawali/src/main/java/org/rajawali3d/scene/Scene.java
+++ b/rajawali/src/main/java/org/rajawali3d/scene/Scene.java
@@ -827,23 +827,14 @@ public class Scene {
 	 * @return {@code boolean} True if the clear task was queued successfully.
 	 */
 	public boolean setSkybox(int resourceId) throws TextureException {
-	        final AFrameTask task = new AFrameTask() {
-	            @Override
-	            protected void doTask() {
-			for (int i = 0, j = mCameras.size(); i < j; ++i)
-				mCameras.get(i).setFarPlane(1000);
-	            }
-	        };
-		synchronized (mNextSkyboxLock) {
-			mNextSkybox = new Cube(700, true, false);
-			mNextSkybox.setDoubleSided(true);
-			mSkyboxTexture = new Texture("skybox", resourceId);
-			Material material = new Material();
-			material.setColorInfluence(0);
-			material.addTexture(mSkyboxTexture);
-			mNextSkybox.setMaterial(material);
-		}
-        	return internalOfferTask(task);
+		Cube skybox = new Cube(700, true, false);
+		skybox.setDoubleSided(true);
+		Texture texture = new Texture("skybox", resourceId);
+		Material material = new Material();
+		material.setColorInfluence(0);
+		material.addTexture(texture);
+		skybox.setMaterial(material);
+        	return setSkybox(skybox, texture);
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


contriteobserver:
1) Meant reflections or environment mapping of objects, not cameras, sorry.
2) Was trying to leave existing practice (700 and 1000) intact, while addressing #1959 for those in need. 
3) Using median diameter, not media radius.

rpicolet:
1) I won't pretend to understand theses use cases or their issues, I believe you.

2) Oh, I missed that you intended the new setSkybox(Cube, ATexture) as a new API.  I see how that can work, and even though it makes the app responsible for texture construction, it can be used to get the job done.  Did you consider instead using a setSkyboxSize() method (that replaces the underlying Cube object) that could work with the existing setSkybox()/updateSkybox() methods, and/or adding set/update wrappers with a size parameter?  Any reason not to add something similar later on (e.g. in 2.0)?

3) Also missed that you switched variable/name to "diameter" after computing the two radii, and so could skip the divide by 2. OTOH I'm guessing the extra distance provided by adding the internal radius solves granularity issues with perspective division at large depths?  I always wondered why it was set all the way out to 1000 for a skybox of 700 with an outer radius (of 350 time root 3 =) 610 or so...  

Thanks again for all your user support!

contriteobserver:
Settled on the setSkyBox(Cube, ATexture) for a variety of reasons:
- addresses issues in #1904 
- provides high level API that satisfies most users, and a low level API for those that need it
- locates all skybox creation to a single function (later raised in #1959)
- allows environment maps and skybox to share the same texture.

rpicolet:
Got it, makes sense, thanks for the insights.

TBog:
Can we also have this method?
`	/**`
`	 * Set the skybox using the provided Cube.`
`	 * It is expected to have a material set and the material to have at least a texture.`
`	 *`
`	 * @return {@code boolean} True if the provided cube passes the requirements`
`	 */`
`	public boolean setSkybox( Cube cube )`
`	{`
`		if ( cube == null )`
`			return false;`
`		Material mat = cube.getMaterial();`
`		if ( mat == null || mat.getTextureList().isEmpty() )`
`			return false;`
`		synchronized ( mNextSkyboxLock )`
`		{`
`			mNextSkybox = cube;`
`			mSkyboxTexture = mat.getTextureList().get( 0 );`
`		}`
`		return true;`
`	}`

contriteobserver:
Part of the intent was to funnel all the `setSkybox()` methods through a single function. Why is `setSkyBox(final Cube skybox, final ATexture texture)` problematic for setting custom materials?

contriteobserver:
Part of the original intent was to funnel all the Skybox texture updates through a single method. If we are to support custom materials for Skyboxes, it seems best to make it a first class feature for all the texture update methods. That way, a custom material would survive texture updates. Thoughts?

jwoolston:
Skyboxes as they are now basically only support diffuse textures. I can think of situations where more than this would be desired, but I think in the situation people should be in their own with updates as it will be rare. To this end I think all the texture methods applying to the normal material is fine if we just add a method which takes a material as an argument.

contriteobserver:
We also need to think about Environment maps, so custom skybox materials are properly reflected. It will take a little while to postulate how they can be effectively associated.

jwoolston:
Good point. That far exceeds the scope of this pr so I will merge. Those issues can be addressed later.

PR #1918, opened 2017-06-17
added more inverse matrix unit tests

contriteobserver:
verifies that given an invertable matrix:
  - a matrix times it's inverse equals the identity matrix
  - an inverse times it's matrix equals the identity matrix


jwoolston:
I need to update the 2.0 development branch to pull all your commits in so they don't get lost.

Issue #1920, opened 2017-06-17
Unexpected Environmental Cube Mapping


contriteobserver:
#### Rajawali Version or Branch
master

#### Summary
given six Cube Map face images like this:
![faces](https://user-images.githubusercontent.com/17471201/27255642-3b85a84e-5357-11e7-92f9-a19aad5ce2c5.png)

On a `Sphere(1,32,32)`, one would expect environmental mapping to look (front and side views) like this:
![expected](https://user-images.githubusercontent.com/17471201/27255644-452f525a-5357-11e7-9459-5413f9152ff1.png)

unexpectedly, the environmental mapping looks like this:
![result](https://user-images.githubusercontent.com/17471201/27255648-4eba68c8-5357-11e7-8bec-883113d61e6d.png)

#### Steps to Reproduce
```
        @Override
        protected void initScene() {
            try {
                int[] faces = {
                        R.drawable.face1,
                        R.drawable.face2,
                        R.drawable.face3,
                        R.drawable.face4,
                        R.drawable.face5,
                        R.drawable.face6
                };
                CubeMapTexture texture = new CubeMapTexture("env", faces);
                texture.isEnvironmentTexture(true);
                Material material = new Material();
                material.addTexture(texture);
                material.setColorInfluence(0);

                obj = new Sphere(1,32,32);
                obj.setMaterial(material);
                getCurrentScene().addChild(obj);

                //getCurrentCamera().setPosition(0,0,3*Math.sqrt(3));
                getCurrentCamera().setPosition(3*Math.sqrt(3),0,0);
                getCurrentCamera().setLookAt(obj.getPosition());
```

contriteobserver:
sorry, it is cube maps that needs to camera rotation fix.

PR #1921, opened 2017-06-17
Improved Environmental Cube Mapping

contriteobserver:
made the InverseViewMatrix available to the Environmental Mapping Shader,
(as the inverse function is not yet available in GLSL ES)
and updated the Environmental shader fragment to update mapping with the camera position.

Not 100% perfect, not yet handling camera rotations about the Z axis, but addresses most of #1920

### DIFF
--- a/rajawali/src/main/java/org/rajawali3d/Object3D.java
+++ b/rajawali/src/main/java/org/rajawali3d/Object3D.java
@@ -53,6 +53,7 @@ public class Object3D extends ATransformable3D implements Comparable<Object3D>, 
 	protected final Matrix4 mMVPMatrix = new Matrix4();
 
 	protected final Matrix4 mMVMatrix = new Matrix4();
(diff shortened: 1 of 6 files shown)


rajawali/src/main/java/org/rajawali3d/Object3D.java
```diff
@@ -214,6 +215,10 @@ public void render(Camera camera, final Matrix4 vpMatrix, final Matrix4 projMatr
 		boolean modelMatrixWasRecalculated = onRecalculateModelMatrix(parentMatrix);
 		// -- calculate model view matrix;
 		mMVMatrix.setAll(vMatrix).multiply(mMMatrix);
+		// -- calculate inverse view matrix;
+                Matrix4 inverseView = vMatrix.clone();
```
jwoolston: Instead of cloning here, call `mInverseViewMatrix.setAll(vMatrix).inverse()`

Issue #1923, opened 2017-06-19
Unexpected Shader Fragment Assignment behavior


contriteobserver:
#### Rajawali Version or Branch
master

#### Summary
Given the following fragment:
```
            private class Vec4AssignTestShaderFragment extends AShader implements IShaderFragment
            {
                public final static String SHADER_ID = "VEC4_ASSIGNMENT_TEST";

                public Vec4AssignTestShaderFragment()
                {
                    super(ShaderType.FRAGMENT_SHADER_FRAGMENT);
                }

                @Override
                public void main() {
                    RVec2 test2 = new RVec2("test2");
                    test2.assign(0);
                    test2.x().assign(0);
                    test2.y().assign(0);

                    RVec3 test3 = new RVec3("test3");
                    test3.assign(0);
                    test3.x().assign(0);
                    test3.y().assign(0);
                    test3.z().assign(0);

                    RVec4 test4 = new RVec4("test4");
                    test4.assign(0);
                    test4.w().assign(0);
                    test4.x().assign(0);
                    test4.y().assign(0);
                    test4.z().assign(0);
                }

                @Override
                public String getShaderId() {
                    return SHADER_ID;
                }

                @Override
                public Material.PluginInsertLocation getInsertLocation() {
                    return Material.PluginInsertLocation.IGNORE;
                }

                @Override
                public void bindTextures(int nextIndex) {}
                @Override
                public void unbindTextures() {}
            }
```
one would expect the following GLSL lines to be generated:
```
vec2 test2 = vec2(0.0);
test2.x = 0.0;
test2.y = 0.0;

vec3 test3 = vec3(0.0);
test3.x = 0.0;
test3.y = 0.0;
test3.z = 0.0;

vec4 test4 = vec4(0.0);
test4.w = 0.0;
test4.x = 0.0;
test4.y = 0.0;
test4.z = 0.0;
```
instead we get:
```
vec2 test2 = 0.0;
test2.x = 0.0;
test2.y = 0.0;

vec3 test3 = vec3(0.0);
test3.x = vec3(0.0);
test3.y = vec3(0.0);
test3.z = 0.0;

vec4 test4 = vec3(0.0);
test4.w = 0.0;
test4.x = vec3(0.0);
test4.y = vec3(0.0);
test4.z = 0.0;
```

Issue #1922, opened 2017-06-19
How Can I translate this GLSL Snippet into Rajawali


MartinRGB:
#### Rajawali Version or Branch
Rajawali Examples

#### Device and Android Version
Xiaomi Mi6 & API 25

#### Summary

I wanna write GLSL code in Material Plugin(for example,CustomMaterialPlugin.java) so that the CPU handle functions.And the GPU handle rendering。

Don't know how to deal with things like "xyww"，"fract"，"assign an RVec4 value ","How to show fragment Code in logcat" in Rajawali Shader.

It seemed that there is no guideline of Shader Programming in Rajawali Wiki.

#### Steps to Reproduce
Code in GLSL Like this:
```
 float noise( vec2 _st )
{
    //  https://github.com/BrianSharpe/Wombat/blob/master/Value2D.glsl

    //	establish our grid cell and unit position
    vec2 Pi = floor(_st);
    vec2 Pf = _st - Pi;

    //	calculate the hash.
    vec4 Pt = vec4( Pi.xy, Pi.xy + 1.0 );
    Pt = Pt - floor(Pt * ( 1.0 / 71.0 )) * 71.0;
    Pt += vec2( 26.0, 161.0 ).xyxy;
    Pt *= Pt;
    Pt = Pt.xzxz * Pt.yyww;
    vec4 hash = fract( Pt * ( 1.0 / 951.135664 ) );

    //	blend the results and return
    vec2 blend = Pf * Pf * Pf * (Pf * (Pf * 6.0 - 15.0) + 10.0);
    vec4 blend2 = vec4( blend, vec2( 1.0 - blend ) );
    return dot( hash, blend2.zxzx * blend2.wwyy );
}


#define NUM_OCTAVES 5

float fbm ( in vec2 _st) {
    float v = 0.0;
    float a = 0.5;
    vec2 shift = vec2(100.0);
    // Rotate to reduce axial bias
    mat2 rot = mat2(cos(0.5), sin(0.5),-sin(0.5), cos(0.50));
    for (int i = 0; i < NUM_OCTAVES; ++i) {
        v += a * noise(_st);
        _st = rot * _st * 2.0 + shift;
        a *= 0.5;
    }
    return v;
}

void main(void) {
    vec2 st = gl_FragCoord.xy/u_resolution.x*3. ;
    // st += st * abs(sin(u_time*0.1)*3.0);
    vec3 color = vec3(0.0);

    vec2 q = vec2(0.);
[...]

PR #1924, opened 2017-06-19
corrected Vec2, Vec3 and Vec4 assignments

contriteobserver:
addresses issue #1923 which also cropped up in #1922

### DIFF
--- a/rajawali/src/main/java/org/rajawali3d/materials/shaders/AShaderBase.java
+++ b/rajawali/src/main/java/org/rajawali3d/materials/shaders/AShaderBase.java
@@ -294,7 +294,7 @@ public abstract class AShaderBase {
 		
 		public ShaderVar x()
 		{
-			ShaderVar v = getReturnTypeForOperation(mDataType, mDataType);
+			ShaderVar v = new RFloat();
 			v.setName(this.mName + ".x");
 			v.mInitialized = true;
 			return v;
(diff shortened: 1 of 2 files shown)


contriteobserver:
added unit tests for RVec2, RVec3, RVec4

rajawali/src/main/java/org/rajawali3d/materials/shaders/AShaderBase.java
```diff
@@ -294,31 +294,31 @@ public ShaderVar xy()
...
-			ShaderVar v = getReturnTypeForOperation(mDataType, mDataType);
+			ShaderVar v = new RFloat();
 			v.setName(this.mName + ".y");
 			v.mInitialized = true;
 			return v;
 		}
 		
 		public ShaderVar s()
 		{
-			ShaderVar v = getReturnTypeForOperation(mDataType, mDataType);
```
contriteobserver: I am confused, aside from float, what other data types would have a meaningful assign to x?

contriteobserver:
added RVec2.assign(float, float) and RVec4.assign(float, float, float, float)

rajawali/src/main/java/org/rajawali3d/materials/shaders/AShaderBase.java
jwoolston: something along the lines of `someVec4.x = 1` or `someVec4.y = 0xAB`

Issue #1928, opened 2017-06-21
Unexpected Bezier Curve Tangents


contriteobserver:
#### Rajawali Version or Branch
master

#### Steps to Reproduce
given the following curve, one would expect a tangent at t=0.5 to be `0.6,0.8,0`
```
            Vector3 p0 = new Vector3(0,0,0);
            Vector3 p1 = new Vector3(0,4,0);
            Vector3 p2 = new Vector3(3,4,0);

            Vector3 point = new Vector3();
            QuadraticBezierCurve3D curve = new QuadraticBezierCurve3D();
            curve.addPoint(p0, p1, p2);
            curve.setCalculateTangents(true);
            curve.calculatePoint(point, 0.5);
            Vector3 result = curve.getCurrentTangent();
```
unexpectedly, result is `-0.6,-0.8,0`

contriteobserver:
The problem is also manifest in the `CubicBezierCurve3D.getCurrentTangent()` as in the following:
```
            Vector3 p0 = new Vector3(0,0,0);
            Vector3 p1 = new Vector3(1,0,0);
            Vector3 p2 = new Vector3(1,2,0);
            Vector3 p3 = new Vector3(1,2,3);

            Vector3 point = new Vector3();
            CubicBezierCurve3D curve = new CubicBezierCurve3D();
            curve.addPoint(p0, p1, p2, p3);
            curve.setCalculateTangents(true);
            curve.calculatePoint(point, 0.5);
            Vector3 tangent = curve.getCurrentTangent();
```

PR #1929, opened 2017-06-22
fixed tangent calculations

contriteobserver:
addresses issue #1928 

for Camden Hells, we can drop `QuadraticBezierCurve3D.setCalculateTangents()` and just get a tangent with `QuadraticBezierCurve3D.getTangent(double t)`

### DIFF
--- a/rajawali/src/main/java/org/rajawali3d/curves/QuadraticBezierCurve3D.java
+++ b/rajawali/src/main/java/org/rajawali3d/curves/QuadraticBezierCurve3D.java
@@ -27,14 +27,13 @@ public class QuadraticBezierCurve3D implements ICurve3D {
 	private Vector3 mTmpPoint3;
 	private Vector3 mTempPointNext=new Vector3();
 	
-	private boolean mCalculateTangents;
-	private Vector3 mCurrentTangent;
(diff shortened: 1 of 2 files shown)


rajawali/src/main/java/org/rajawali3d/curves/QuadraticBezierCurve3D.java
```diff
@@ -60,35 +59,27 @@ public void addPoint(Vector3 point1, Vector3 controlPoint, Vector3 point2) {
...
 		mTmpPoint3.setAll(mPoint2);
-		mTmpPoint3.multiply(t * t);
+		mTmpPoint3.multiply(mCurrent * mCurrent);
 		mTmpPoint2.add(mTmpPoint3);
 		result.addAndSet(mTmpPoint1, mTmpPoint2);
 	}
 
 	public Vector3 getCurrentTangent() {
-		return mCurrentTangent;
+		Vector3 start_portion = mControlPoint.clone().subtract(mPoint1).multiply(1-mCurrent);
```
jwoolston: Calling clone each time like this will cause any animation using this method to generate a good deal of garbage. Can you make it use scratch members instead?

PR #1934, opened 2017-06-24
fixed camera Z axis rotation issue

contriteobserver:
forgot to transpose the reflection matrix in my last checkin

resolves the rest of #1920

### DIFF
--- a/rajawali/src/main/java/org/rajawali3d/Object3D.java
+++ b/rajawali/src/main/java/org/rajawali3d/Object3D.java
@@ -216,7 +216,7 @@ public class Object3D extends ATransformable3D implements Comparable<Object3D>, 
 		// -- calculate model view matrix;
 		mMVMatrix.setAll(vMatrix).multiply(mMMatrix);
 		// -- calculate inverse view matrix;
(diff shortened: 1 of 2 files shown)


jwoolston:
@contriteobserver What is the purpose of transposing the inverse view matrix? That seems suspect to me. Also, what is the reason for removing the normalization?

contriteobserver:
the transpose changes the order of the reflection operations so the scaling is applied correctly, the normalization got the scaling right, but the order was wrong, once I saw it I removed the normalization in favour of the transpose.

rami-dabain:
is this related? https://github.com/Rajawali/Rajawali/issues/1936

contriteobserver:
@rami-dabain this issue is unrelated to #1936, the title was an attempt at a simple description of a complex concept, namely that the reflection looked odd when the geometry emphasised the effects of non-uniform scaling, i.e. when the camera was rotated on multiple axes. The second normalisation vs transpose is really a uniform vs non-uniform scaling issue. It is easy to see when you know what you are looking for, but the visual effect is difficult to describe.

PR #1950, opened 2017-07-16
first pass Octree unit tests

contriteobserver:


jwoolston:
@contriteobserver The implementation I wrote of Octree's in Rajawali 1.x is garbage, and we will be scrapping it in 2.x in favor for a clean right. I appreciate all your hard work but don't waste any time on this one.

Issue #1966, opened 2017-08-21
Add StreamingTexture and normal Texture to an Plane?


yeyezizhi:
#### Rajawali Version or Branch
org.rajawali3d:rajawali:1.1.777@aar

#### Device and Android Version
huawei honor 6, android 6.0

#### Summary
I want to add an audio to a plane and add another image texture at the same time. I have successfully added the StreamingTexture to the plane, but the app cashes after i added another image texture. So i want to know whether i can add two texture to the same object.  Any advice would be appreciated!

contriteobserver:
In 1.1.777 it is possible to add two video or two image textures to a material.
However, there is an issue when adding an image and video texture to a material.
/pull request #1967 addresses the mapping of mixed (image and video) textures.

yeyezizhi:
Thanks for your reply! I'll try it.

PR #1967, opened 2017-08-23
added support of mixed (Image and Video) texture mapping.

contriteobserver:
addresses issue #1966


jwoolston:
I just noticed this myself the other day. Awesome!

Issue #896, opened 2013-07-01
Obj parser fixes


MasDennis:
We get a lot of question & issues about this (#858 for instance). The Obj parser seems to be the most used parser so I think these issues have to be fixed for the Anchor Steam release.

Here's a list of what should be fixed:
- [ ] support per face textures by creating texture atlasses
- [ ] support 'g' groups. It currently reads groups but it doesn't handle them properly
- [ ] more feedback to the user through `RajLog`. So we have better information when the user reports issues.
- [ ] throw an error when .mtl file couldn't be found. At the moment it only writes a log message which most users miss.
- [ ] throw an error when a texture cannot be found.
- [ ] look into the specular settings in the .mtl file. Some files have a weird specular factor which tends to make the model white. What is the standard?

Issue #1619, opened 2016-03-18
Material and Texture managers fail to properly handle multiple renderers.


jwoolston:
#### Rajawali Version or Branch
`master`

#### Summary
Per the discusions of #1256 and #1518:

If there is more than one rendered registered, as can happen in a ViewPager, then the last registered renderer wins.
A ViewPager typically loads the visible page, and then the immediately adjacent pages.
The off-screen pages are registered last, resulting in all texture and material tasks getting targeted to that off-screen renderer; the visible on-screen rendered never gets the textures/materials.

I think Rajawali's intention is that there is only one active renderer at a time.
The renderer is typically activated in onResume and deactivated in onPause.
However, even the non-active renderers need to load textures.

It seems to me that the basis of the TextureManager and MaterialManager is flawed.
The intention seems to be that multiple renderers can efficiently add duplicate textures and materials to the respective manager, and a single instance of those textures/materials will be saved.
The flaw is that those textures/materials are only added to the currently active renderer, when they should be available to all renderers.

Issue #1468, opened 2015-04-09
AWD Skeleton animation has visual artifacting


jwoolston:
This could be caused by #1465, or it could have some other cause such as missing data in the file. Further investigation is required.

Issue #1465, opened 2015-04-09
AWD Mirroring fix broke AWD Skeleton parsing


jwoolston:

jwoolston:
The conversion to right handed coordinates from left handed coordinates is required. I will have to do some digging in AWD to see if there are other implications related to the skeleton caused by their left handed system.

jwoolston:
Further testing confirms that the right handed conversion is not perfect, some rotations are still mirrored.

PR #1981, opened 2017-09-14
Update REQUIREMENTS.md

jwoolston:


dev/camdenhells/REQUIREMENTS.md
```diff
@@ -44,7 +44,7 @@ For reviews and updates:
     * Does not imply support for the entirety of that version's API
     * Does imply that engine features available in earlier versions are still available
   * Using an engine feature that depends on a version greater than is available on a device:
-    * Will in general throw a runtime error exception
+    * Will in general throw a checked exception, forcing the user code to deal with it.
```
rpicolet: I would prefer to avoid checked exceptions except when it is an absolutely necessary part of the API contract, i.e. for something that cannot be otherwise reliably tested/predicted. I believe this approach reduces client code clutter, and discourages both client use of exceptions for normal control flow and the client just re-throwing all Exceptions to avoid the try/catch code when there are numerous checked exceptions that rarely occur, or that never occur if the contract is observed.

In this case, that would suggest that we should either require that the client checks versions before using a version-specific feature (as called out in the next bullet), or throw a checked exception, but not both, and since it is predictable, I would say the former. 

When we resolve this, we should probably add a section to the coding conventions, as this is a general issue.

dev/camdenhells/REQUIREMENTS.md
```diff
@@ -55,34 +55,37 @@ For reviews and updates:
...
-2. Allow multiple `GlSurfaceViews` per Activity ? [2.1]
-  * Is this actually worthwhile?
-  * Only one render thread/context/Renderer can exist at once, so SurfaceViews are mutually exclusive 
-    * E.g. `ViewPager` support (per #1619)
-    * Is timing of render context switch still based on window attachment?
-3. Wallpaper surface view multisampling fix (#1559)
-4. Provide query/configuration constants (bit flags) for all nVidia GL ES extensions (#1742) [2.1]
-  * Do any other vendors or extension categories need this?
+2. Decouple Android views from the EGL context to allow offline rendering/initialization
+3. ~Allow multiple `GlSurfaceViews` per Activity ? [2.1]~
```
rpicolet: If we can't or won't support multiple GlSurfaceViews per Activity (presumably with only one unpaused/running at a time), then should #1619 be declined? How else can there be multiple renderers?

dev/camdenhells/REQUIREMENTS.md
```diff
@@ -55,34 +55,37 @@ For reviews and updates:
...
 1. Provide (major and minor) version query for the current GL ES render context
-2. Allow multiple `GlSurfaceViews` per Activity ? [2.1]
-  * Is this actually worthwhile?
-  * Only one render thread/context/Renderer can exist at once, so SurfaceViews are mutually exclusive 
-    * E.g. `ViewPager` support (per #1619)
-    * Is timing of render context switch still based on window attachment?
-3. Wallpaper surface view multisampling fix (#1559)
-4. Provide query/configuration constants (bit flags) for all nVidia GL ES extensions (#1742) [2.1]
-  * Do any other vendors or extension categories need this?
+2. Decouple Android views from the EGL context to allow offline rendering/initialization
```
rpicolet: I'm not at all sure what offline rendering means, and it seems like EGL contexts must be tightly coupled to GlSurfaceView/GlSurfaceTextureView (and so GLThread) implementations.  Is the real goal here just to support off-screen rendering?

dev/camdenhells/REQUIREMENTS.md
```diff
@@ -55,34 +55,37 @@ For reviews and updates:
...
 
 ### 3.2.1. Model files
  
-1. Enable "artist control" e.g. material specs, transparency/blending... ?
-2. OBJ fixes/improvements (#896 - multiple items!)
-3. Assimp support ? [2.1]
-4. Improved AWD support/bug fixes ? [2.1]  (#1185, #1465, #1468)
-5. MD5 child naming (#1454) [2.1]
+1. ~Enable "artist control" e.g. material specs, transparency/blending... ?~ Proper loader support (#2) will take care of this.
+2. Assimp support
```
rpicolet: Does this mean CH will no longer support OBJ, AWD, or MD5? I.e. should we close issues #896, #1185, #1465, #1468, and #1454 if Assimp is the new go-to model file format?

dev/camdenhells/REQUIREMENTS.md
```diff
@@ -105,15 +108,16 @@ Provide separate specification of a rendered view for a scene model. Each scene
 2. One camera, switchable, and a dynamic set of any number/mix of lights
   * No inherent geometry/vertex data
   * Embedded in the scene, transformable 
-  * World coordinates can be querie
+  * World coordinates can be queried
   * "Containers", optional visualization geometries for debugging or app functions can be added
   * Switch betwen whole light sets at once ? [2.1]
-3. Skybox texture/bitmaps and size, and/or background color; all mutable
+3. Skybox texture/bitmaps and size, and/or background color and/or materials; all mutable
```
rpicolet: Why are materials added separately here?  This bullet should have been just about defining the background of the scene view, i.e. a skybox and/or a clear color, but I threw in the Skybox parameters as well. 
We should probably have a separate  requirement for any skybox enhancements per se,

Are you using "material" to mean "program"? If so, I want to advocate for consistently distinguishing throughout CH between material specs and programs (some of which implement material/lighting specs). Or do you mean that skyboxes should support material/lighting specs in addition to textures (i.e. "skybox texture/bitmaps/_material_spec and size, and/or background color")?  If so, are the skybox material spec and texture/bitmaps mutually exclusive, and is the material spec per cube face? 

BTW,  I'd like to cleanly separate a Skybox child from the Cube primitive. So if the goal is to allow material specs then we should probably modify the Cube instead to accept a material (face array?)...

dev/camdenhells/REQUIREMENTS.md
```diff
@@ -173,7 +177,7 @@ Provide separate specification of a rendered view for a scene model. Each scene
   * Any number of frame renders to readable off-screen image buffers
 4. Multiple render targets per framebuffer (#1862)
   * E.g. for deferred lighting and ambient occlusion g-buffers
-  * All render targets are same size
+  * Different size attachments are supported per render target
```
rpicolet: Do you mean that different render targets in a framebuffer can be backed by attached buffers of different sizes as long as they all are at least of the shared target size? I.e. "All render targets share the same render area size in their attached buffers"? 

Or do you mean there is a use case for different size render areas in simultaneously-attached color/depth-stencil buffers?

dev/camdenhells/REQUIREMENTS.md
```diff
@@ -295,8 +295,9 @@ This does not preclude performance impacts; clearly there will be limits, but ha
   * No public API fields; provide accessors
   * Document implicit default thread assumptions 
     * E.g. render thread-only (not safe) unless otherwise specified
-  * Mark explicit thread-safety per type and/or per method (method overrides type)
+  * Mark explicit thread-safety per type and/or per method with annotation
   * App responsible for observing thread safety markers, no built-in enforcement
+   * Consider creating Android lint plugin
```
rpicolet: [2.1] ?

dev/camdenhells/REQUIREMENTS.md
```diff
@@ -328,7 +329,7 @@ This does not preclude performance impacts; clearly there will be limits, but ha
   * Isolated - from all (other) render thread/event processing
     * Frame draws, window re-sizes, queued events, other CRUD transactions, etc.
     * Implies running on the render thread as an event, queued from any thread
-    * TODO how to handle queued transactions after window detach/loss of context...
+    * Queued transactions after window detach/loss of context are dropped
```
rpicolet: Right, so then it is purely the client's responsibility to handle this situation if/as needed.

dev/camdenhells/REQUIREMENTS.md
jwoolston: Assimp is a library which supports something like 40 different formats, OBJ and MD5 among them. I have not completely written off AWD either, but it has become a lower priority. I want to keep the tickets open to make sure the new implementation is free of those bugs.

dev/camdenhells/REQUIREMENTS.md
jwoolston: - `ObjectRenderer` is what I meant
- I'm fine with a Skybox/Skysphere primitive
- I need to do some research before I can make a decision on the per face textures.

dev/camdenhells/REQUIREMENTS.md
jwoolston: There is a use case for different resolution attachments to a single render buffer. Sometimes you want a computer texture which due to what you are computing doesn't require full resolution, or on the flip side requires higher resolution.

jwoolston: Scratch this. I don't know what I was thinking. The mechanism for multiple attachments can't possibly support this

dev/camdenhells/REQUIREMENTS.md
jwoolston: Anything future

dev/camdenhells/REQUIREMENTS.md
jwoolston: Perhaps I should have been clearer. I wish to throw checked exceptions for things that are not catostophic. For example, texture operations. Nothing about a texture operation failing should prevent attempting to fix the problem. For this reason, the programmer should be forced at compile time to make a decision to either gracefully degrade/recover or if they are being lazy, bury the exception and deal with whatever the render behavior becomes.

For this where no recovery is possible, I don't wish to throw checked exceptions.

dev/camdenhells/REQUIREMENTS.md
jwoolston: The nature of what EGL is does not require a true user facing window. Some implementations do and I'm not yet sure if Android is among them. If you use the built in classes of Android then yes, it's required but we already don't use the Android ones, so we could potentially have a shot at this. 

Yes the goal is to support off screen renders. For example, I originally got into rajawali because I was working on a weather app that wished to render an atmospheric shader that calculated Raleigh scattering. This doesn't change much per second or even. Per minute so the goal was to basically render an image every so often, and then use that image in an image view. This never happened but it is this kind of situation I wish to support

dev/camdenhells/REQUIREMENTS.md
jwoolston: After some thought I think I misunderstood this. I will update.

dev/camdenhells/REQUIREMENTS.md
rpicolet: Cool use case, didn't think of that.  Essentially one-shot on-demand rendering to a bitmap. Should be orthogonal to current GlSurfaceView/GlThread usage, no? Seems like a 2.1 req?

dev/camdenhells/REQUIREMENTS.md
jwoolston: Yes

dev/camdenhells/REQUIREMENTS.md
rpicolet: Ah, not just another model file format. Makes sense now.  So we would basically throw out most of the current Raj parsers and use assimp to populate object data/scenes?

dev/camdenhells/REQUIREMENTS.md
jwoolston: I have never been happy with the implementations of rajawali's surface/texture views. From the day I added them to the library I havn't trusted them completely. The surface view uses Android's under the hood and the texture view is basically a clone of what the surface view does. My distrust largely comes from how complex the classes are. I would like to examine the prospects of factoring out the GL thread. There is another reason for this. The GL thread runs in the view and we run a Rajawali thread on top of that currently. Really all this rajawali thread does is tell the GL thread to render a frame, rather than using the built in mechanisms for rendering at a certain frame rate or continuously. This is super wasteful. I am not sure to what extent we can accomplish this but it is definately a major shift in how things operate and therefore a key 2.0 requirement in my mind.

dev/camdenhells/REQUIREMENTS.md
jwoolston: Correct. My original plan was to interface Assimp through an API in a similar spirit to the current Loader API. Custom formats or other formats not supported in Assimp could be implemented through this API (like they are now). AWD would be a good example. Of course, adding them to Assimp is always a possibility too, but that means C++ code so personnel to do that is limited.

dev/camdenhells/REQUIREMENTS.md
rpicolet: Yep to the ObjectRenderer for skyboxes...  Exactly what I had in mind too.  I'd probably move the skybox/sphere out of the primitives package and into the sceneview package, as they make more sense to me as an integral part of a SceneView rather than as general primitives, and they track the Camera location, etc.   Not worried about per face stuff for now.

dev/camdenhells/REQUIREMENTS.md
rpicolet: Ah, assimp is native, so I'm guessing we'll need a JNI/JNA wrapper between the new loader API you're suggesting and assimp proper?

jwoolston: Correct. A JNI wrapper is also required to access many vendor specific extensions so I have already begun the work that will make this possible.

dev/camdenhells/REQUIREMENTS.md
rpicolet: I don't think Android's GlSurfaceView/GlThread offers actual frame timing, only continuous vs on-request modes, so I don't know how we could get away from using a separate timer/thread/executor or whatever to generate the render-frame requests. And not sure how we could get around it even if we wrote our own render thread, as it still has to handle surface size change, loss of context, and queued events, but maybe there's a way...

dev/camdenhells/REQUIREMENTS.md
contriteobserver: When messing with skybox/sphere we also need to keep Environment Mapping in mind. It is always embarrassing when reflective materials do not reflect the sky.

jwoolston: Are you suggesting that when a Skybox is active any environment map materials should Automatically update?

dev/camdenhells/REQUIREMENTS.md
rpicolet: Actually, after thinking about it, it should be possible and relatively straightforward to support on-demand off-screen rendering without significant architectural changes such as decoupling from SurfaceViews. If an app is already using a SurfaceView, it should be easy to add a single frame off-screen render whenever needed; if the app is not using a Surface view, it should be easy to just create a hidden instance and use it to do the off-screen render.

jwoolston: Correct, but I would like this to be a fallback solution. If at all possible I would like to better control the egl context creation.

dev/camdenhells/REQUIREMENTS.md
```diff
@@ -178,7 +178,9 @@ Provide separate specification of a rendered view for a scene model. Each scene
   * Any number of frame renders to readable off-screen image buffers
 4. Multiple render targets per framebuffer (#1862)
   * E.g. for deferred lighting and ambient occlusion g-buffers
-  * Different size attachments are supported per render target
+  * All attachments per render target are the same size (resolution).
+  * Different storage formats per attachment are supported per render target.
```
rpicolet: I'm guessing you mean "are supported per framebuffer", i.e. the attached color buffers can be of different internal formats?

rpicolet:
Not sure how to add new issue in-place in the doc, so putting it here...

For 3.2.4 Scene Views, bullet 1, I'd like to change the Scene to be mutable, as that seems like it might come in useful for RecyclerView type behaviors, i.e. having a fixed number of scrollable PIP SceneViews that can present any number of streamed-in Scenes.

jwoolston:
I'm fine with mutable scenes

dev/camdenhells/REQUIREMENTS.md
jwoolston: Correct

rpicolet:
OK here's a cleanup list before we merge the reqs:

### 3.2.2 Colors & Textures
delete Help! 
2. New functions - delete

### 3.3.4 Effects
* Fog ?
* Blend ?

### 3.4.1 CPU/GPU 
3. RenderScript - s/b [2.1] ?

### 3.4.2 Memory Resources 
1. Minimize allocations... 
* Render target images/framebuffer attachments s/b per render context, not per frame render or render pass
* If there are no shared resources to manage per scene or scene view, then let's say so or remove those scopes

### 3.6.1 Enable render thread safety
Method annotation overrides type annotation?
Lint plugin s/b [2.1]; also covered below, so could be deleted

### 3.6.3 Transactions
3. Callback - UI thread-queueing variant s/b [2.0] (no "?")

# 4. Testing & Debugging
1. Unit Tests
* I'd still like to see some details/guidelines about what categories of classes and methods will clearly benefit from unit tests, especially if we also invest in internal runtime assertions (next item)
2. Internal runtime debug logging/tracing/assertions 
* S/b [2.0] (no "?")
* All debug logging, tracing & assertions should be removed from production code, whether by Proguard or other means
3. Compile-time API usage
* Lint plugin again - s/b [2.1]
4. Runtime API usage
* S/b [2.0] - uses same mechanism as internal runtime assertions
5 Debug visualizations
* Are these all [2.0] ? 

# 5. Modularity
* Separate modules for add-ons - s/b [2.1] ?
  * Add Font-to-texture and gesture support modules
* All-in-one debug variant still [2.0]

# 6. User Docs
* The TBD lists for the examples app and the Wiki guidelines should include at least the bare minimums we are willing to ship with/commit to now.

dev/camdenhells/REQUIREMENTS.md
contriteobserver: Currently, we have cube and sphere environment mapping. If we are going to add an equirectangular mapped skyshere geometry, we should also add equirectangular environment mapping so we can continue to share textures between both sky geometry and environment mapping.

dev/camdenhells/REQUIREMENTS.md
contriteobserver: Have we considered moving object data/scenes loading into a separate module?
I would prefer to keep native dependencies out of the core.

jwoolston: That was already planned, though not spelled out. I will update it. The core will only contain the java interface that all loaders will access/implement. This ultimately applies to the extensions too.

rpicolet:
I stuck some more pegs in the ground just to have a clear 2.0 scope.  The area I am least confident about is unit testing, that could use a good critique for sure.  Otherwise, feel free to change 2,0 to 2.1 or the other way round if you disagree with my pegs. One last review/update should do it I think, then we can merge and move on with clear goals.

dev/camdenhells/REQUIREMENTS.md
```diff
@@ -324,41 +350,75 @@ This does not preclude performance impacts; clearly there will be limits, but ha
...
+# 4. Testing, Validation, and Debugging Requirements
+
+## 4.1 Compile-time
+
+  * Consistently specify any constraints on all member and parameter values
+    * Specify all references as either nullable or non-null
+    * Specify integer and float ranges whenever less than built-in language limits
+  * ~Validate public API-usage [2.1]~
+     * ~Create an Android lint plugin to detect invocations of non API methods?~
+     * Seems like a reliable library-specific tool would be a lot of work/maintenance, and unlikely that client developers would download/install/use it anyway
```
jwoolston: My intention was to make it run in a gradle task, forcefully, but I agree its going to be a lot of work and not the top priority.

dev/camdenhells/REQUIREMENTS.md
```diff
@@ -324,41 +350,75 @@ This does not preclude performance impacts; clearly there will be limits, but ha
...
+
+## 4.2 Build-time 
+
+  Help! I'm trying to define the line on what should be unit tested and what should not, and don't even understand the hardware/software environment in which unit tests run (i.e. emulators use what version of OpenGL? is there an Android process/activity context? etc)
```
jwoolston: There are two kinds of unit tests. Those that exist in a java context only (such as the tests for the math package - Matrix4, etc.), and those that exist in a GL context (such as VBO tests). At present, the GL tests are setup around using an Android activity and surfaceview, but this may change if we create our own GL thread. The tests only need a valid GL context, they don't need to display. 

At the moment, the travis build can only execute the non-GL unit tests due to a limitation of the emulator combined with travis. Essentially because it is a headless emulator, GL support is not included. Hardware tests are currently executed on my private device farm, with occasional runs by @ToxicBakery to supplement the available devices. Additional users would be welcomed. We can add any of these devices to the list in the `v2.0-development` `README.md`.

As for what should be unit tested, anything that can be black boxed. Essentially, if you can expect a certain output for a given input because it interacts only with itself, then it is a candidate for unit testing. Minimal interaction is a candidate for integration testing - which I have been adding as well. Scene graphs are a good example of this.

dev/camdenhells/REQUIREMENTS.md
```diff
@@ -206,15 +228,15 @@ Provide separate specification of a rendered view for a scene model. Each scene
...
-* Fog [?]
-* Blur [?]
-* Blend [?]
-* Bloom [?]
-* FXAA [?]
-* SSAA [2.1]
-* Sepia [?]
-* Scanline [?]
-* Vignette [?]
+* Fog [2.1]
```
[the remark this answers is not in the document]
rpicolet: Good, I was just wondering why you left these two unresolved in your edits, so I figured this would do it.

dev/camdenhells/REQUIREMENTS.md
rpicolet: Yeah, it doesn't seem like it will be straightforward for client devs to keep a lint tool in syncnwith the library...

dev/camdenhells/REQUIREMENTS.md
rpicolet: Thanks, that helps a lot.  Will update accordingly.

dev/camdenhells/REQUIREMENTS.md
```diff
@@ -324,41 +350,75 @@ This does not preclude performance impacts; clearly there will be limits, but ha
...
+## 4.2 Build-time 
+
+  Help! I'm trying to define the line on what should be unit tested and what should not, and don't even understand the hardware/software environment in which unit tests run (i.e. emulators use what version of OpenGL? is there an Android process/activity context? etc)
+  
+  * Automated unit/integration test harness and tests 
+  * For all public API and internal interface methods except those:
+    * With significant historical runtime event dependencies or behavioral states, and/or
+    * Whose success cannot be (at least partially) reasonably determined as value assertions
+  * Includes e.g. contructors, simple get/set accessors, deterministic functions/transforms
+  * Excludes e.g. frame draws, client callbacks
```
jwoolston: My ultimate goal is that the examples can have full integration testing where screenshots can be captured and compared to known good screenshots from the same device. This will represent a full stack integration test that should help us catch when examples break as well as assert that low level changes did not have unforeseen larger scale effects such as what #1986 is resolving.

rpicolet: WIll add examples integration tests (as [2.1] unless you think [2.0]). Should we mention instrumentation? Seems like we are currently not using it, but maybe I'm missing something in my analysis.  There is also a new Android Test Orchestrator that looks like it might be useful for this...  All new to me, just exploring the edges for requirements purposes for now.

jwoolston: 2.1 is fine for now. I will look into the Test Orchestrator

dev/camdenhells/REQUIREMENTS.md
```diff
@@ -324,41 +350,75 @@ This does not preclude performance impacts; clearly there will be limits, but ha
...
+## 4.3. Run-time
+
+For engine and app development and integration testing, and for debugging support issues
+
+1. Provide unified logging
+  * The usual levels - verbose, debug, info, warning, error, wtf
+  * All messages should identify the class and method
+  * Core engine/rendering component (scenes, scene view, render pass, etc.) messages should also identify the instance
+  * All warning and error messages should include specific causes and data values if appropriate
+  * All error messages should be accompanied by a thrown exception
```
jwoolston: Unless rendering can continue. For example, a singular program failing to compile is definitely an error, however rendering can continue with other programs. Only parts of the render using the failed program will not display.

dev/camdenhells/REQUIREMENTS.md
```diff
@@ -324,41 +350,75 @@ This does not preclude performance impacts; clearly there will be limits, but ha
...
+2. Provide debug-level method tracing messages
```
jwoolston: I would say this should be verbose level. Its very useful to have but I would want it enabled only when there was a particular problem i wished to track. I would prefer to leave the debug level clear for end user development.

dev/camdenhells/REQUIREMENTS.md
```diff
@@ -324,41 +350,75 @@ This does not preclude performance impacts; clearly there will be limits, but ha
...
+  * The usual levels - verbose, debug, info, warning, error, wtf
+  * All messages should identify the class and method
+  * Core engine/rendering component (scenes, scene view, render pass, etc.) messages should also identify the instance
+  * All warning and error messages should include specific causes and data values if appropriate
+  * All error messages should be accompanied by a thrown exception
+2. Provide debug-level method tracing messages 
+  * For key lifecycle event (create/init/start/run/pause/stop/destroy etc.) handlers of engine/rendering components
+    * Includes all active components: scenes, scene views, render passes, etc.
+    * Does not include passive model objects, textures, resources 
+  * Tracing dynamically enabled per instance; disabled by default
```
jwoolston: Can you clarify what you mean by this?

dev/camdenhells/REQUIREMENTS.md
```diff
@@ -324,41 +350,75 @@ This does not preclude performance impacts; clearly there will be limits, but ha
...
+  * All error messages should be accompanied by a thrown exception
+2. Provide debug-level method tracing messages 
+  * For key lifecycle event (create/init/start/run/pause/stop/destroy etc.) handlers of engine/rendering components
+    * Includes all active components: scenes, scene views, render passes, etc.
+    * Does not include passive model objects, textures, resources 
+  * Tracing dynamically enabled per instance; disabled by default  
+3. Use logging and trace messaging judiciously to avoid excessive performance impacts or major frame delays
+  * Comment out or delete developmental log messages when no longer needed
+  * Avoid frame-rate messages
+    * Accumulate/integrate/summarize per-frame data, and log e.g. once per second or so
```
jwoolston: There has to be a library out there for this already - I don't want to go down the road of implementing this ourselves.

rpicolet: Badly worded on my part, I meant "for example, accumulate..." ; If such data were going to be gathered. Didn't mean to require it, will fix

dev/camdenhells/REQUIREMENTS.md
```diff
@@ -324,41 +350,75 @@ This does not preclude performance impacts; clearly there will be limits, but ha
...
+4. API and internal interface usage validation, e.g. debug-only assertions/exceptions
+  * Check input parameters for all public API methods and all internal interface methods 
+  * Check system and instance state for lifecycle event handlers
+  * Check thread for non-thread-safe methods
+  * Except for error-level log messages, strip all logging, tracing, and assertion code from release builds
+    * Remove code rather than simply using conditional checks
```
jwoolston: Proguard can be used for this task.

dev/camdenhells/REQUIREMENTS.md
```diff
@@ -324,41 +350,75 @@ This does not preclude performance impacts; clearly there will be limits, but ha
...
+      * Modifiers, animations, rendering effects
+      * Font-to-texture
+      * Gestures
+      * Wear
+      * TV?
       * Wallpaper
-  * Still provide all-in-one module?
-  * Publish debug/development version in addition to release version of each module?
+  * Always provide all-in-one module in addition to core plus add-on modules
+  * Publish debug/development version in addition to release version of each module
```
jwoolston: All snapshots are debug/development versions. Only tagged versions will be built as true releases (stripping unnecessary code, etc).

dev/camdenhells/REQUIREMENTS.md
rpicolet: I mean there would be e.g. a simple boolean for each core control/rendering component that determines whether any trace messages are actually logged for that instance; the trace methods for an instance would all be conditional. So you could just enable tracing on this scene, or that scene view, or that render pass. Most of the time there would be no trace output.

dev/camdenhells/REQUIREMENTS.md
rpicolet: Verbose is fine with me, will change.

dev/camdenhells/REQUIREMENTS.md
rpicolet: Doesn't that imply some fairly sophisticated runtime logic to decide if rendering can continue and how to partially disable the bad parts (e.g. this object has a bad program/VBO/whatever so skip rendering it)?  I.e. should we use error to mean unrecoverable, and warning to mean we think we can keep going (without any extra logic)?

jwoolston: Its pretty straight forward and we have them in place now (1.x). If the VBO handles are invalid or the program handles are invalid, we skip that render call and move to the next. I am also ok with the idea of this being a warning and keeping the errors accompanied by exceptions

rpicolet: OK, will add the caveat for GL resource warnings and leave errors with exceptions

dev/camdenhells/REQUIREMENTS.md
rpicolet: Ok, review to make sure I got your drift...

Issue #1980, opened 2017-09-14
Why scene turns upside down when apply EffectsPass


peterrichardc:
Why my scene turns upside down when apply FXAAPass[Anti-Aliasing]? 

Im setting FXAAPass in my scene, but after apply my scene turns upside down...

    @Override
    protected void initScene() {
        sceneRoom = new Scene(this);

        DirectionalLight mDirectionalLight = new DirectionalLight(1f, .2f, -1.0f);
        mDirectionalLight.setColor(1.0f, 1.0f, 1.0f);
        mDirectionalLight.setPower(2);
        sceneRoom.addLight(mDirectionalLight);

        //Add Objects.
        addRoomObj();
        addBedObj();

        //Set Camera Pos.
        sceneRoom.getCamera().setPosition(0, 8, 30);
        replaceAndSwitchScene(getCurrentScene(), sceneRoom);

        //Anti analising <<<<<<<<<<<<<<<<<<<<<<<<<<<<< HERE \/
        postProcessingManager = new PostProcessingManager(this);
        RenderPass renderPass = new RenderPass(sceneRoom, sceneRoom.getCamera(), 0);
        postProcessingManager.addPass(renderPass);

        EffectPass antialiasPass = new FXAAPass();
        antialiasPass.setRenderToScreen(true);
        postProcessingManager.addPass(antialiasPass);

    }

    @Override
    protected void onRender(long ellapsedRealtime, double deltaTime) {
        //super.onRender(ellapsedRealtime, deltaTime);

        postProcessingManager.render(ellapsedRealtime, deltaTime);
    }

#### Rajawali Version or Branch
1.1.970.

#### Device and Android Version
6.0

contriteobserver:
Cannot seem to reproduce this issue. What is wrong with this test code:
```
        @Override
        protected void initScene() {
            Scene scene = new Scene(this);

            scene.setBackgroundColor(Color.BLUE);

            DirectionalLight key = new DirectionalLight(-3,-4,-5);
            key.setPower(2);
            scene.addLight(key);

            LoaderOBJ loader = new LoaderOBJ(getResources(), getTextureManager(), R.raw.up01_obj);
            try {
                loader.parse();
                obj = loader.getParsedObject();
                obj.setScale(0.3f);
                scene.addChild(obj);
            } catch (ParsingException e) {
                e.printStackTrace();
            }

            scene.getCamera().setPosition(3,3,3);
[...]

PR #1986, opened 2017-09-26
ScreenQuad now handles CW and CCW texture mapping

contriteobserver:
ScreenQuad texture mapping Fix for #1980

### DIFF
--- a/rajawali/src/androidTest/java/org/rajawali3d/primities/ScreenQuadTest.java
+++ b/rajawali/src/androidTest/java/org/rajawali3d/primities/ScreenQuadTest.java
@@ -1,0 +1,28 @@
+package org.rajawali3d;
+
+import static org.junit.Assert.*;
+import org.junit.*;
+
+import org.rajawali3d.primitives.ScreenQuad;
+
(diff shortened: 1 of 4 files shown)


rajawali/src/androidTest/java/org/rajawali3d/primities/ScreenQuadTest.java
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    @After
+    public void teardown() throws Exception {
+        screenQuad = null;
+    }
+
+    @Test
+    public void testConstructor() throws Exception {
+        assertNotNull(screenQuad);
```
ToxicBakery: Seems a little redundant. If a constructor fails it throws a checked exception otherwise the allocation happens therefore the setup function would fail for every test function.

contriteobserver: Working on a better test fixture, but I didn't want the test fixture work to hold up getting the ScreenQuad fix reviewed.

PR #1995, opened 2017-10-09
V2.0 core dev

rpicolet:
WIP for render control architecture and logging/debugging utiliies


core/src/main/java/c/org/rajawali3d/logging/LoggingComponent.java
```diff
@@ -0,0 +1,383 @@
...
+import org.rajawali3d.BuildConfig;
+import timber.log.Timber;
+
+/**
+ * Instance-specific debug-mode and release-mode logging and debug utilities
+ *
+ * @author Randy Picolet
+ */
+
+public abstract class LoggingComponent {
```
jwoolston: Rather than rolling our own here, I think it makes the most sense to utilize `Timber`. It has already been added as a dependency.

rpicolet: Um, I did use Timber...   The purpose of this class is to add capabilities beyond static logging (more/better context for each message, plus asserts and configurable method tracing), but underneath everything there are the two Timber trees - DebugLog and ReleaseLog, whichever one is planted will handle all the messages generated in LoggingComponent...  I chose to wrap Timber rather than have Timber dependencies spread through the engine.  If we hit any problems with Timber, we can then replace it very easily.  Am i missing your concern?

jwoolston: No I missed the usage in all of the PR. My appologies.

rpicolet: Not a problem.  There is a lot to absorb, all feedback is great, and I have no doubt I will take my turns with missing stuff.

core/src/main/java/c/org/rajawali3d/control/RenderSurfaceView.java
```diff
@@ -0,0 +1,37 @@
...
+    /**
+     * Request rendering of the next complete frame for the SurfaceView.
+     */
+    void requestFrameRender();
+
+    /**
+     *
+     * @param runnable
+     */
+    void queueEvent(Runnable runnable);
```
jwoolston: Can you please elaborate on the reasoning for both `queueEvent` and `post`?

core/src/main/java/c/org/rajawali3d/object/renderers/ObjectRendererImpl.java
```diff
@@ -171,9 +171,14 @@ private void applyDepth() {
     }
 
     private void checkDoubleSided(@NonNull ObjectRenderer lastUsed) {
-        if (isDoubleSided && !lastUsed.isDoubleSided()) {
-            // We are double sided, the last renderer is not - disable culling
-            GLES20.glDisable(GLES20.GL_CULL_FACE);
+        if (isDoubleSided) {
```
jwoolston: Why was this expanded into more verbose logic?

core/src/main/java/c/org/rajawali3d/object/renderers/ObjectRendererImpl.java
```diff
@@ -192,28 +197,44 @@ private void checkDoubleSided(@NonNull ObjectRenderer lastUsed) {
...
 
     private void checkBlending(@NonNull ObjectRenderer lastUsed) {
-        if (isBlended && !lastUsed.isBlended()) {
-            // We are blended, the last renderer is not - enable
-            GLES20.glEnable(GLES20.GL_BLEND);
-            GLES20.glBlendFunc(blendSourceFactor, blendDestinationFactor);
-        } else if (lastUsed.isBlended()) {
-            // We are not blended, the last renderer is - disable
-            GLES20.glDisable(GLES20.GL_BLEND);
+        if (isBlended) {
```
jwoolston: Why was this expanded into more verbose logic?

core/src/main/java/c/org/rajawali3d/control/RenderSurfaceView.java
```diff
@@ -0,0 +1,37 @@
...
+import c.org.rajawali3d.surface.SurfaceView;
+
+/**
+ * Internal extensions of {@link SurfaceView} required by a {@link RenderControl} implementation. The goal is to
+ * define a minimal interface that can reasonably be expected to be supportable by current (GLES) and future
+ * (Vulkan?) graphics systems.
+ *
+ * @author Randy Picolet
+ */
+public interface RenderSurfaceView extends SurfaceView {
```
jwoolston: Why not just fuse this class with `SurfaceView`?

core/src/main/java/c/org/rajawali3d/object/renderers/ObjectRendererImpl.java
rpicolet: If isDoubleSided is true, but the lastUsed was also, then the else clause would be used...

core/src/main/java/c/org/rajawali3d/object/renderers/ObjectRendererImpl.java
rpicolet: If  isBlended is true and lastUsed was also, then the else clause will be used.

core/src/main/java/c/org/rajawali3d/object/renderers/ObjectRendererImpl.java
jwoolston: But in that situation, there is nothing to do regarding double sided, which is why the clause was omitted originally.

core/src/main/java/c/org/rajawali3d/object/renderers/ObjectRendererImpl.java
jwoolston: Same as above

core/src/main/java/c/org/rajawali3d/object/renderers/ObjectRendererImpl.java
rpicolet: The original else clause assumed _not_ double sided and changed GL, so if the condition fails because both the current and last used are doubleSided, then GL will be configured for non-double-sided...

core/src/main/java/c/org/rajawali3d/object/renderers/ObjectRendererImpl.java
rpicolet: Same for me too

core/src/main/java/c/org/rajawali3d/control/RenderSurfaceView.java
rpicolet: There is actually no reason it should extend SurfaceView, which is the public client API. This is an internal API that defines what CoreControl needs from a surface view implementation.  I will remove the extends clause.

core/src/main/java/c/org/rajawali3d/control/RenderSurfaceView.java
rpicolet: queueEvent() posts to the render thread, post() to the main/UI thread

jwoolston: I would prefer method names which reflected that.

rpicolet: [Not sure if my edit was before or after your comment, so moving foreard I will stop editing except for spelling and such, not semantics or content]

Since the methods are already present in both GlesSurfaceView and GlesSurfaceTextureView, I just used them as-is so those classes did not have to be changed or introduce another delegation, but it makes sense looking forward to more surface view impls e.g. for Vulkan.  I will change to `queueToRenderThread()` and `queueToMainThread()`.  Good call.

jwoolston: Sounds good. Also, with our rather loose discussion about possibly extracting the render thread class, we basically will be keeping our own version of both TextureView and SurfaceView, so we can ultimately call them whatever we want.

core/src/main/java/c/org/rajawali3d/object/renderers/ObjectRendererImpl.java
rpicolet: Should have said "The top-level else clause assumes _not_ double sided and changesGL..."

core/src/main/java/c/org/rajawali3d/control/RenderSurfaceView.java
rpicolet: I'm adding a PublicApi annotation to help with this since we need to clearly mark API anyway for everyone's good...

jwoolston: I would rather annote private APIs. Generally I want to try and protect these apis as much as possible with visibility, but in cases where that is not possible, we can annotate accordingly. Our public API is likely to be much larger than the truly private API, so its easier to maintain the private listing.

core/src/main/java/c/org/rajawali3d/control/gles/GlesRenderControl.java
```diff
@@ -0,0 +1,10 @@
+package c.org.rajawali3d.control.gles;
+
+import c.org.rajawali3d.control.RenderControl;
+
+/**
+ * @author Randy Picolet
+ */
+
+public interface GlesRenderControl extends RenderControl {
```
jwoolston: I know that this additional interface level is specifically to handle Vulkan vs GLES, but you should document that so it is clear.

core/src/main/java/c/org/rajawali3d/surface/gles/BaseGlesSurfaceRenderer.java
```diff
@@ -1,30 +1,35 @@
...
- * Shared implementation for GL ES extensions of {@link ARenderControl}
+ * Shared implementation for GLES {@link SurfaceRenderer} extensions of {@link CoreControl}
  *
  * @author Randy Picolet
  */
-abstract class GLESRenderer extends ARenderControl {
 
-    GLESRenderer(Context context, RenderSurfaceView renderSurfaceView,
-                 RenderControlClient renderControlClient, double initialFrameRate) {
+abstract class BaseGlesSurfaceRenderer extends CoreControl implements GlesRenderControl {
```
jwoolston: `CoreControl` implements `RenderControl` while `GlesRenderControl` extends it. We need to avoid this cycle.

core/src/main/java/c/org/rajawali3d/surface/gles/BaseGlesSurfaceRenderer.java
```diff
@@ -1,30 +1,35 @@
...
+    BaseGlesSurfaceRenderer(Context context, RenderSurfaceView renderSurfaceView,
+                            RenderControlClient renderControlClient, double initialFrameRate) {
         super(context, renderSurfaceView, renderControlClient, initialFrameRate);
     }
 
     /**
-     * Callback to notify that the EGL context has been acquired and a valid GL thread with EGL surface now exists.
+     *
      */
     protected void onRenderContextAcquired() {
```
jwoolston: Don't omit `@Override` annotations.

jwoolston: I know this is from old code, but the call to `Capabilities.getInstance()` here is redundant as we have established it in the `configure()` method of the surfaces.

core/src/main/java/c/org/rajawali3d/control/BaseControl.java
```diff
@@ -0,0 +1,41 @@
...
+    //
+
+    /**
+     * Hook method required by all {@link ControlComponent}s, called from {@link CoreControl#onRenderFrame()};
+     * runs any tasks queued for this component
+     *
+     * @param deltaTime
+     */
+    @RenderThread
+    protected abstract void onFrameStart(@FloatRange(from = 0.0) final double deltaTime);
```
jwoolston: I would like to retain the total elapsed time as well.

core/src/main/java/c/org/rajawali3d/control/RenderSurfaceView.java
rpicolet: Using (I assume `public`) visibility consistently seems problematic, as there are numerous places where the engine needs to use public across its own packages, and numerous places where the client may need to exploit package/protected visibility.  

But I'm ok with flagging internal types & methods versus external APIs, as long as we can tag a whole type rather than every method. Then, if we reliably maintain separate internal/external interfaces (i.e. don't allow individual interface _methods_ to be tagged), and clearly specify that overridden methods inherit their API-ness (so we don't have to maintain tags all the way down the implementation/override tree), then we only have to also annotate internal methods that are not implementing an internal interface but are included in an non-internal class...  

For example, we could just tag CoreControl as internal at the class level, and methods in CoreControl that implement e.g. the external RenderControl interface would not need to be tagged either.  The tags are just documentation, so as long as the rules for their interpretation are also documented, then when a support issue comes in we can just say "sorry that's internal use only" and point them to the appropriate documentation.

Not sure about how to tag callbacks.  They are technically internal, even though the client has to implement them...  Maybe they just don't get tagged period, since all of that is obvious.

core/src/main/java/c/org/rajawali3d/control/gles/GlesRenderControl.java
rpicolet: Forgot this existed. I'm not actually sure its needed, and I'm really sure it should not extend `RenderControl`...

core/src/main/java/c/org/rajawali3d/surface/gles/BaseGlesSurfaceRenderer.java
rpicolet: Agreed, its wrong will fix.

core/src/main/java/c/org/rajawali3d/surface/gles/BaseGlesSurfaceRenderer.java
rpicolet: Never on purpose, always flag em for me.

rpicolet: I'm missing where else the Capabilities singleton gets created in the startup sequence...  OTOH, I'm not sure what is gained by instantiating it if it is never used, so maybe this call to getInstance() should just be deleted?

core/src/main/java/c/org/rajawali3d/control/BaseControl.java
rpicolet: I think we had a discussion about this in the past, when talking about animations; the elapsed time is always available from `RenderStatus.getFramesElapsedTime()` when needed.  I don't think we use/need it internally.

core/src/main/java/c/org/rajawali3d/control/gles/GlesRenderControl.java
jwoolston: If it doesn't serve a purpose I'm fine with nuking it. One less interface to track

core/src/main/java/c/org/rajawali3d/surface/gles/BaseGlesSurfaceRenderer.java
jwoolston: My point was to remove it. It used to be used for populating the capabilities once a render context existed but I have changed it to no longer require that.

rpicolet: Gone

core/src/main/java/c/org/rajawali3d/control/gles/GlesRenderControl.java
rpicolet: Gone

core/src/main/java/c/org/rajawali3d/surface/gles/BaseGlesSurfaceRenderer.java
rpicolet: Although in this case its just the shared implementation, not overriding anything...  Will document

jwoolston: What do you mean by shared implementation?

core/src/main/java/c/org/rajawali3d/control/RenderSurfaceView.java
jwoolston: Not sure I fully understand what you are saying. Skip the annotation for now, we can decide on it at a future point.

rpicolet: OK for now.  I will just have to read my own comments (instead of my own tags) to avoid mixing internal/external interfaces as I did here...

core/src/main/java/c/org/rajawali3d/surface/gles/BaseGlesSurfaceRenderer.java
rpicolet: This method is used by both GL surface view renderers to get the context version number before forwarding up to `CoreControl.onRenderContextAcquired()`. Didn't want to duplicate the code in both places is all.

jwoolston: The problem is it implements an interface which defines a function of that name. Intentional or not, you overrode it.

rpicolet: The interface method is public and has context type/version parameters, this one is protected and has none. Different signatures. This one wraps/calls the interface method on behalf of both surface views, since they both figure out the version the same way; it does not override or extend the interface method.

jwoolston: Then I would argue one of them needs a different name.

rpicolet: That's legit, renamed it to `notifyRenderContextAcquired`.

core/src/main/java/c/org/rajawali3d/annotations/InternalUseOnly.java
```diff
@@ -0,0 +1,22 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ *  Flag a type or method for internal/engine use only, and specifically not supported for use by engine clients as
+ *  part of its external API
+ *
+ *  Flagging a method applies to all of its implementations/extensions/overrides; flagging a type implies flagging
```
jwoolston: What does `flagging a type implies flagging all methods in that type` mean?

core/src/main/java/c/org/rajawali3d/annotations/RequiresRenderTask.java
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * Flag a method that should run only in the context of a {@link RenderTask} or a similar
+ * context, i.e. on the RenderThread, and isolated from any other render events including frame renders
+ *
+ * @author Randy Picolet
+ */
+@Documented
+@Retention(RetentionPolicy.CLASS)
+@Target({ElementType.METHOD})
+public @interface RequiresRenderTask {
```
jwoolston: Off the cuff, I am concerned this may be confusing without much aid. I will reserve final judgment until I get through the entirety of this PR. In the meantime, this comment will serve as a reminder to myself to re-evaluate.

core/src/main/java/c/org/rajawali3d/control/RenderContextType.java
```diff
@@ -0,0 +1,16 @@
+package c.org.rajawali3d.control;
+
+/**
+ * Basic render context types
+ *
+ * @author Randy Picolet
+ */
+public enum RenderContextType {
```
jwoolston: Can we make this an android `@IntDef annotation instead?

core/src/main/java/c/org/rajawali3d/control/RenderControl.java
```diff
@@ -0,0 +1,212 @@
...
+     *
+     * @return {@code true} if the calling thread is the render thread.
+     */
+    boolean isRenderThread();
+
+    /**
+     *
+     * @param runnable
+     */
+    void queueToMainThread(Runnable runnable);
```
jwoolston: `runnable` should be annotated as `@NonNull`

core/src/main/java/c/org/rajawali3d/control/RenderControl.java
```diff
@@ -0,0 +1,212 @@
...
+
+    /**
+     *
+     * @param runnable
+     */
+    void queueToRenderThread(Runnable runnable);
```
jwoolston: `runnable` should be annotated as `@NonNull`

core/src/main/java/c/org/rajawali3d/control/RenderControl.java
```diff
@@ -0,0 +1,212 @@
...
+     * {@link Scene#initialize()}, and should take any needed actions.
+     *
+     * @param scene A {@link Scene} to add; duplicates are quietly ignored.
+     */
+    @RequiresRenderTask
+    void addScene(@NonNull Scene scene);
+
+    /**
+     * Removes a {@link Scene} from the list of Scene frame delegates. The {@link Scene} will be notified via
+     * {@link Scene#destroy()}, and should take any needed actions.
```
jwoolston: I know I'm responsible for this but I think we should consider changing this contract to be `Scene#removed()` instead of implying that removal equals destruction.

core/src/main/java/c/org/rajawali3d/control/RenderControl.java
```diff
@@ -0,0 +1,212 @@
...
+     * </p>
+     *
+     * @param sceneView The {@link SceneView} to add; duplicates are quietly ignored.
+     */
+    @RequiresRenderTask
+    void insertSceneView(@NonNull SceneView sceneView, int depthOrder);
+
+    /**
+     * Removes a {@link SceneView} from the list of SceneView frame delegates. The {@link SceneView} will be notified
+     * via {@link SceneView#destroy()}, and should take any needed actions.
```
jwoolston: Same contract issue as with `addScene()`

core/src/main/java/c/org/rajawali3d/scene/SceneModifier.java
```diff
@@ -1,21 +0,0 @@
...
-
-import android.support.annotation.NonNull;
-import c.org.rajawali3d.annotations.RequiresWriteLock;
-
-/**
- * Interface which must be implemented in order to modify a {@link Scene}.
- *
- * @author Jared Woolston (Jared.Woolston@gmail.com)
- */
-public interface SceneModifier {
```
jwoolston: Can you help me track what this turned into?

core/src/main/java/c/org/rajawali3d/scene/graph/FlatTree.java
```diff
@@ -77,7 +77,6 @@ public void recalculateBoundsForAdd(@NonNull SceneNode added) {
 
     @RequiresReadLock
     @NonNull
-    @Override
```
jwoolston: Why was this annotation removed?

core/src/main/java/c/org/rajawali3d/annotations/InternalUseOnly.java
rpicolet: It means tagging at the class or interface level means all the methods in that type are also internal only, so we don't have to tag every method in an internal-only class or interface.  At your suggestion, I haven't used this anywhere.  Instead/for now, I started using "Internal" in the actual name, which certainly makes the point more visible.

core/src/main/java/c/org/rajawali3d/annotations/RequiresRenderTask.java
rpicolet: The point is to identify methods that should be called only in a render transaction. 

In my next update, I've renamed the `@RenderTask` annotation to `@RenderTransaction`, and this one to `@RequiresRenderTransaction`, as I co-opted the `RenderTask` class name from its old use in 1.x for start-of-frame `Renderer`-queued tasks to the new render-thread-queued tasks that run completely outside of draw frame events.  So in general, `@RenderTransaction` flags a client method that runs on the render thread but outside frame updates., so e.g. callbacks to clients running sequentially under a context creation, surface size change, or loss of context event control flow are `@RenderTransaction`s, in addition to the new `RenderTask` that allows apps to do whatever they need to when they need to.

core/src/main/java/c/org/rajawali3d/control/RenderContextType.java
rpicolet: Makes sense, sure.

core/src/main/java/c/org/rajawali3d/control/RenderControl.java
rpicolet: Yep will do

core/src/main/java/c/org/rajawali3d/control/RenderControl.java
rpicolet: Yep will do

core/src/main/java/c/org/rajawali3d/control/RenderControl.java
rpicolet: I'm not sure the callback is even necessary. I'm leaning towards adding/removing being internal operations that do the GL resource management under the covers, and letting the client do whatever they want before or after adding or removing.  In other words, `addToRenderControl()` means allocate/configure whatever GL resources are necessary internally to start using the Scene at the next frame draw, and `removeFromRenderControl()` means release whatever resources a Scene is currently using because it will not longer participate in frame draws.  That's why `onAddToRenderControl()` is part of the internal interface, the client shouldn't have to do anything

jwoolston: Ok, let's see what this fleshes out to.

core/src/main/java/c/org/rajawali3d/control/RenderControl.java
rpicolet: And same response...

core/src/main/java/c/org/rajawali3d/scene/SceneModifier.java
rpicolet: Nothing.  Or rather, whatever the client would have wanted to do under this should be done in a RenderTask, along with any changes to its dependent SceneViews or to other Scenes/Sceneviews as needed by their app's logic.  All part of getting away from fine-grained updates and providing atomic transactions of any granularity

core/src/main/java/c/org/rajawali3d/control/RenderControl.java
rpicolet: Yeah, it's taken me a while to disentangle the old single-scene model lifecycle semantics from my brain, but I think its about there. 

Of course, I've said similar things before...

rpicolet:
Can’t find this comment on GitHub, replying via email

I removed the @Override from FlatTree.intersect() because I removed that method from the SceneInternal interface and there was nothing left to override. i removed the method from the interface because I added the visibleObjectIntersection method intended to pre-filter NodeParents and invisible/hidden Object3Ds from the list handed to a SceneView and which may be iterated mutiple times and to simplify the rendering logic a bit.

Issue #1184, opened 2014-02-05
ColorAnimation3D strange behaviour with CycleInterpolator


semenoh:
I'm highlighting selected object with ColorAnimation3D like this:

``` Java
mPicker.setOnObjectPickedListener(new OnObjectPickedListener() {
            @Override
            public void onObjectPicked(Object3D object) {
                ColorAnimation3D anm= new ColorAnimation3D(Color.WHITE, Color.parseColor("#80000000")); 

                anm.setDuration(1000);
                anm.setRepeatCount(0);
                anm.setRepeatMode(RepeatMode.NONE);
                anm.setInterpolator(new CycleInterpolator(1));
                anm.setTransformable3D(object);

                registerAnimation(anm);

                anm.play();
            }
        });
}
```

I works well but after animation is finished approximately half a second later it changes again to some color from range I set in animation. BTW i depends on time delay,  so it seams to me that it is some interpolated value.

PR #2094, opened 2019-01-15
addresses issue #1184

contriteobserver:
in Animation.update(), mElapsedtime could be set larger than mDuration,
causing a value greater than one to be sent to the Interpolator.


ToxicBakery:
Looks good, make sure you update your fork so future PRs can be built. :+1:

PR #2114, opened 2019-01-24
Use recent VR

hannesa2:
Current master uses a very old GoogleVR lib, it's copied and not used by dependencies.
This is a one way usage, so I tried to use most recent one.

This PR is based on branch https://github.com/Rajawali/Rajawali/tree/feature/vuforia-8 , but with my knowledge I was stopped here.

image

Maybe someone knows more. I just want to share my try and if you see this is a one way, you can decline/close this PR


ToxicBakery:
I think I'm missing some context, was there a reason for doing this off of the vuforia branch? I don't want to merge that branch until I get the sample working again so if this can be done cleanly off develop that would definitely be a better approach imo.

hannesa2:
> I think I'm missing some context, was there a reason for doing this off of the vuforia branch? 

The reason was this: because the vuforia branch start on my side in Emultor

> I don't want to merge that branch until I get the sample working again so if this can be done cleanly off develop that would definitely be a better approach imo.

Sure, I would not merge it too. It was just to ask, if someone knows a solution

ToxicBakery:
I think that was a miscommunication. I'm completely fine with your changes I just wanted them to be on a clean branch. I'm going to reopen this and I'll take a look asap but I'm pretty tied up until at least Sunday.

goldenheartpl:
Anyone developing this actually?

hannesa2:
I at least rebased branch, to solve conflicts, but I guess it needs a change to CardsBoard [source](https://github.com/googlevr/gvr-android-sdk) 


image

hannesa2:
@goldenheartpl but you can jump into it and make it work !

PR #2132, opened 2019-09-05
fix CircleCI

hannesa2:
It's still broken https://circleci.com/gh/Rajawali/Rajawali/109 as I see https://hub.docker.com/r/rajawali/rajawali/tags
you use now only `1.1` instead with leading _release-_ of previous `release-1.0`

Sorry my fault, I didn't double checked it

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 defaults: &defaults
   working_directory: ~/repo
   docker:
-    - image: rajawali/rajawali:release-1.1
+    - image: rajawali/rajawali:1.1
   environment:
     TERM: dumb
 


ToxicBakery:
definitely my fault, I jacked that one up good lol

hannesa2:
sad, still failing https://circleci.com/gh/Rajawali/Rajawali/110 and local 
`./gradlew :rajawali:dokkaGhPages` works.
I'm out now.

Btw, which "security reason" prevents a CI build of a PR ?

ToxicBakery:
I think historically it was because private variables can be leaked but I don't think circleci has that limitation so it's possible we can fix that now.

PR #2134, opened 2019-10-03
Fbx import

ValkA:
Using these commits I'm able to import an FBX file that was exported by blender 2.8 and converted by [FBX Converter 13.3](https://www.autodesk.com/developer-network/platform-technologies/fbx-converter-archives) (2009-2010 ascii formats seem to work). 

Normals seem to be good. 
NormalsObject3D  can be used to render other object's normals for debugging.

I'm also baking rotation and scale of objects before exporting, just to be on the safe side.
To do that in blender 2.8, select the object in object mode then ctrl+A and then "Rotation & Scale״.

My blender export settings:
image

Big thanks to - https://banexdevblog.wordpress.com/2014/06/23/a-quick-tutorial-about-the-fbx-ascii-format/

### DIFF
--- a/rajawali/build.gradle
+++ b/rajawali/build.gradle
@@ -57,16 +57,16 @@ dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
 }
 
-apply from: '../publish.gradle'
+//apply from: '../publish.gradle'
 apply plugin: 'kotlin-android'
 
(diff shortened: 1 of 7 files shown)


ValkA:
there are a few cases where some of the trigs are not drawn because of culling, I think it because of a triangulation I'm doing there, I'll try to fix that.

ToxicBakery:
Changes look fine but you disabled parts of the build script, can you revert those?

PR #2178, opened 2020-05-20
addressed Kotlin warnings

contriteobserver:
fixed "Type mismatch: inferred type is String? but String was expected" warnings

### DIFF
--- a/rajawali/src/main/java/org/rajawali3d/view/TextureView.kt
+++ b/rajawali/src/main/java/org/rajawali3d/view/TextureView.kt
@@ -936,10 +936,10 @@ open class TextureView @JvmOverloads constructor(
             } catch (e: InterruptedException) {
                 // fall thru and exit normally
             } catch (e: IllegalStateException) {
-                Log.e("RajawaliGLThread", e.message)
+                Log.e("RajawaliGLThread", "" + e.message)
                 showToast(e.message)
             } catch (e: Exception) {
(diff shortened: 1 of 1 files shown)


rajawali/src/main/java/org/rajawali3d/view/TextureView.kt
```diff
@@ -936,10 +936,10 @@ open class TextureView @JvmOverloads constructor(
             } catch (e: InterruptedException) {
                 // fall thru and exit normally
             } catch (e: IllegalStateException) {
-                Log.e("RajawaliGLThread", e.message)
+                Log.e("RajawaliGLThread", "" + e.message)
```
ToxicBakery, line 939: What is the error you tried to solve here? This isn't really kotlin standard so im guessing here that message my be null but log doesn't want a null? If that's the case you can instead just do `e.message ?: ""`

contriteobserver: Not an error, a warning:
```
> Task :rajawali:compileReleaseKotlin
w: /home/elkq/Rajawali/rajawali/src/main/java/org/rajawali3d/view/TextureView.kt: (939, 43): Type mismatch: inferred type is String? but String was expected
w: /home/elkq/Rajawali/rajawali/src/main/java/org/rajawali3d/view/TextureView.kt: (942, 43): Type mismatch: inferred type is String? but String was expected
``` 
will update the patch as mentioned.

PR #2214, opened 2020-11-04
Smoke test with screenshot

hannesa2:
Currently there is only one[ pointless Espresso test](https://github.com/Rajawali/Rajawali/blob/master/rajawali/src/androidTest/java/org/rajawali3d/renderer/RendererTest.java) which tests nothing.

This pull request tests only if build apk is able to start without a crash and additionally makes a screenshot and uploads it to CI job

This **Rajawali-Espresso** Zip contains
image

* logcat output
* Espresso test html result
* Screenshot of test

Give it a try and look into it !

### DIFF
--- a/.github/workflows/Android-CI-Espresso.yml
+++ b/.github/workflows/Android-CI-Espresso.yml
@@ -19,24 +19,28 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
       - name: Install Android SDK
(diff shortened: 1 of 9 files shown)


hannesa2:
It's only test the non-Vuforia flavor, because Vuforia has no `x86` abi for a fast Espresso test

ToxicBakery (approved):

PR #2219, opened 2020-11-08
Rotate example

hannesa2:
I'm investigation the rotation behavior, but it confuses me.

image

That's why I made an example, but something is confusing here !

Change Z axis need a `monkey.rotX = value`

                dataRenderer?.rotateDataX(progress.toDouble())
                textMixed.text = "Z=${progress}"

Change X axis need a `monkey.rotZ = value`

                dataRenderer?.rotateDataZ(progress.toDouble())
                textHorizontal.text = "X=${progress}"

Something is here not right, beginning with my XYZ rotation image, the sample or Rajawali library

This is one of two issues I observe in rotate something with Rajawali

rotateExample

### DIFF
--- a/examples/build.gradle
+++ b/examples/build.gradle
@@ -1,5 +1,6 @@
 apply plugin: 'com.android.application'
 apply plugin: 'kotlin-android'
+apply plugin: 'kotlin-android-extensions'
 
 android {
     compileSdkVersion rootProject.compileSdk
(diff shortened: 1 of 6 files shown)


hannesa2:
To have a more clear picture about the issue I added more logging.
To proper rotate X I'm forced to set **Z** `monkey.rotZ = value` and result is resulting in **X**

image

hannesa2:
For me it looks like a wrong/mistaken implementation of `.rotX` and `.rotZ`

Please explain or confirm

`.rotY` works like expected

PR #2226, opened 2020-11-21
Simplify CI

hannesa2:


hannesa2:
It depends on https://github.com/Rajawali/Rajawali/pull/2224 so I append it there

PR #2272, opened 2021-01-30
cleaning out an unused VertexAnimationObject3D dependency 

contriteobserver:
cleaning out an VertexAnimationObject3D dependency that somehow got into Geometry3D


hannesa2:
Maybe we should run `ktlint` during pull request to check for unused imports.

ToxicBakery (approved):

Issue #2291, opened 2021-02-02
Need float array Specular Color setters


contriteobserver:
#### Rajawali Version or Branch
master

#### Summary
Need float array specular color setters, as we currently provide for `Material.SetColor()`

#### Steps to Reproduce
in Material.java we have both:
```
    public void setColor(int color)
    public void setColor(float[] color)
```

in `Specular.Phong` and `Specular.CookTorrance` we only have:
```
    public void setSpecularColor(int color)
```

PR #2292, opened 2021-02-02
provides a float array Specular Color setter

contriteobserver:
does not change legacy interface specification

addresses #2291

### DIFF
--- a/rajawali/src/main/java/org/rajawali3d/materials/shaders/fragments/specular/PhongFragmentShaderFragment.java
+++ b/rajawali/src/main/java/org/rajawali3d/materials/shaders/fragments/specular/PhongFragmentShaderFragment.java
@@ -33,7 +33,7 @@ public class PhongFragmentShaderFragment extends ATextureFragmentShaderFragment 
 	private RFloat muShininess;
 	private RFloat muSpecularIntensity;
 	
-	private float[] mSpecularColor;
+	private float[] mSpecularColor = {1,1,1};
 	private float mShininess;
 	private float mSpecularIntensity;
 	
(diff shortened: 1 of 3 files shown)


ToxicBakery (approved):

rajawali/src/main/java/org/rajawali3d/materials/shaders/fragments/specular/PhongFragmentShaderFragment.java
```diff
@@ -43,16 +43,13 @@
...
 	}
 	
-	public PhongFragmentShaderFragment(List<ALight> lights, int specularColor, float shininess, float specularIntensity, List<ATexture> textures) {
+	public PhongFragmentShaderFragment(List<ALight> lights, float[] specularColor, float shininess, float specularIntensity, List<ATexture> textures) {
 		super(textures);
-		mSpecularColor = new float[] { 1, 1, 1 };
-		mSpecularColor[0] = (float)Color.red(specularColor) / 255.f;
-		mSpecularColor[1] = (float)Color.green(specularColor) / 255.f;
-		mSpecularColor[2] = (float)Color.blue(specularColor) / 255.f;
+		mSpecularColor = mSpecularColor;
```
aimiebell, line 52: I detect that this code is problematic. According to the [Correctness (CORRECTNESS)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#correctness-correctness), [SA: Self assignment of value (SA_SELF_ASSIGNMENT)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#sa-self-assignment-of-field-sa-field-self-assignment).
This method contains a self assignment of a field; e.g.
Such assignments are useless, and may indicate a logic error or typo.

Issue #2301, opened 2021-02-10
improve object names generated by LoaderOBJ


contriteobserver:
#### Rajawali Version or Branch
Associations between sibling objects generated by `LoaderOBJ` are opaque. 
An improved naming scheme that simplifies the identification of how an object instance relates to the stored model is desired.

Related to issue #2294 

#### Device and Android Version
master

#### Summary
Rajawali implements one object per material, however, models containing many materials are not uncommon. When `LoaderOBJ` encounters an object displaying more than one material, it reconciles this conflict by separating the parent object into children, one for each material. However, identifying how this child geometry is associated with each material is problematic, as each child is given a randomly generated name.

#### Steps to Reproduce
```
        @Override
        protected void initScene() {
            try {
                getCurrentScene().setBackgroundColor(Color.CYAN & Color.GRAY);

                DirectionalLight key = new DirectionalLight(-4,-4,-4);
                key.setPower(5/4f);
                getCurrentScene().addLight(key);

                LoaderOBJ loader = new LoaderOBJ(this, R.raw.posing1doll_obj);
                loader.parse();
                Object3D obj = loader.getParsedObject();
                obj.setMaterial(material);
                getCurrentScene().addChild(obj);

                Log.i(getClass().getSimpleName() + ".InitScreen", "object graph\n" + loader.toString());

                getCurrentCamera().setUpAxis(Vector3.Axis.Y);
                getCurrentCamera().setPosition(3,2,1);
                getCurrentCamera().setLookAt(obj.getPosition().clone().add(0,1,0));
            } catch (Exception e) {
                Log.e(getClass().getSimpleName(), e.getMessage());
            }
        }
```

#### Current Log Output
```
CurrentRenderer.InitScreen: object graph
    -->GROUP default
    	-->Object5521
    	-->Object4497
    	-->Object1506
    	-->Object1356
    	-->Object1444
    	-->Standing_Pose
```

#### Desired Log Output
```
DesiredRenderer.InitScreen: object graph
    -->GROUP default
    	-->Arm.Left3549
    	-->Arm.Right4840
    	-->Leg.Left2670
    	-->Leg.Right4299
    	-->Head1699
    	-->Standing_Pose
```

jonathanchh1:
Excellent. i see that you guys implemented it yesterday. Thats amazing. How can i start utilizing the changes?

contriteobserver:
[...]

PR #2302, opened 2021-02-10
implemented revealing object names

contriteobserver:
by prefixing each generated child with the name of the source material that caused the object to be created.

addresses #2301

### DIFF
--- a/rajawali/src/main/java/org/rajawali3d/loader/LoaderOBJ.java
+++ b/rajawali/src/main/java/org/rajawali3d/loader/LoaderOBJ.java
@@ -284,7 +284,7 @@ public class LoaderOBJ extends AMeshLoader {
 					currentMaterialName = parts.nextToken();
 					if(currentObjHasFaces) {
 						objIndices.add(currObjIndexData);
-						currObjIndexData = new ObjIndexData(new Object3D(generateObjectName()));
+						currObjIndexData = new ObjIndexData(new Object3D(generateObjectName(currentMaterialName)));
 						RajLog.i("Parsing object: " + currObjIndexData.targetObj.getName());
 						addChildSetParent(currentGroup, currObjIndexData.targetObj);
 						currentObjHasFaces = false;
@@ -416,7 +416,12 @@ public class LoaderOBJ extends AMeshLoader {
 	}
 
 	private static String generateObjectName() {
-		return "Object" + (int) (Math.random() * 10000);
+		return generateObjectName(null);
+	}
+
+	private static String generateObjectName(String prefix) {
+		if(prefix==null) prefix = "Object";
+		return prefix + (int) (Math.random() * 9e3 + 1e3);
 	}
 
 	/**
--- a/rajawali/src/main/java/org/rajawali3d/loader/LoaderOBJ.java
+++ b/rajawali/src/main/java/org/rajawali3d/loader/LoaderOBJ.java
@@ -44,6 +44,7 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Stack;
 import java.util.StringTokenizer;
+import java.util.concurrent.atomic.AtomicInteger;
 
 /**
  * The most important thing is that the model should be triangulated. Rajawali doesn�t accept quads, only tris. In Blender, this is an option you can select in the exporter. In a program like MeshLab, this is done automatically.
@@ -96,6 +97,7 @@ public class LoaderOBJ extends AMeshLoader {
     protected final String DIFFUSE_TEX_MAP = "map_Kd";
 
     private boolean mNeedToRenameMtl = true;
+    static AtomicInteger mUniqueInstanceId = new AtomicInteger(1000);
 
     public LoaderOBJ(Renderer renderer, String fileOnSDCard) {
     	super(renderer, fileOnSDCard);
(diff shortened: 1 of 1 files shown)


rajawali/src/main/java/org/rajawali3d/loader/LoaderOBJ.java
```diff
@@ -416,7 +416,12 @@ private void mergeGroupsAsObjects(Object3D object) {
 	}
 
 	private static String generateObjectName() {
-		return "Object" + (int) (Math.random() * 10000);
+		return generateObjectName(null);
+	}
+
+	private static String generateObjectName(String prefix) {
+		if(prefix==null) prefix = "Object";
+		return prefix + (int) (Math.random() * 9e3 + 1e3);
```
hannesa2: Why not using a timestamp (or UUID) ?

contriteobserver: Excellent idea, I went with a timestamp as the ramifications of that system call is better understood.

rajawali/src/main/java/org/rajawali3d/loader/LoaderOBJ.java
```diff
@@ -421,7 +422,7 @@ private static String generateObjectName() {
 
 	private static String generateObjectName(String prefix) {
 		if(prefix==null) prefix = "Object";
-		return prefix + (int) (Math.random() * 9e3 + 1e3);
+		return prefix + Long.toString(SystemClock.elapsedRealtime());
```
hannesa2: Keep in mind, this is in nanos, I would go with millis `System.currentTimeMillis()` which is more common to use as unixtimestamp. But I'm fine with it.

contriteobserver: `System.currentTimeMillis()` would not necessarily provide the necessary resolution, as models sometimes duplicate names, and we can parse multiple child objects in a mS

object graph
    -->GROUP default
    	-->Arm.Left2597618548
    	-->Arm.Right2597618769
    	-->Leg.Left2597619008
    	-->Leg.Right2597619185
    	-->Head2597619359
    	-->Standing_Pose

jonathanchh1: I see you guys have now implemented this feature. Is there away to pull these changes or can it be released? I'm working on something really urgent and i would love to have this feature. Thank you!

hannesa2: Really urgent ? Nothing to do, here you go 
`implementation "com.github.contriteobserver:Rajawali:LoadedMtlObjNaming-SNAPSHOT"`
or
`implementation "com.github.contriteobserver:Rajawali:45cf99dc15"`

With jitpack.io you can have a library of any commit. Please see [details](https://jitpack.io/#contriteobserver/Rajawali/45cf99dc15)

jonathanchh1: Yes, definitely urgent! thank you again!

ToxicBakery: If the object names need guaranteed uniqueness, why not a uuid?

rajawali/src/main/java/org/rajawali3d/loader/LoaderOBJ.java
ToxicBakery: Seeing this now after my other question. I really think uuid implies the intent vs a random number or nanos usage.

rajawali/src/main/java/org/rajawali3d/loader/LoaderOBJ.java
contriteobserver: UUIDs alone are insufficient, as there is a need for name to be revealing. Also, legacy identifiers had the classifier "Object" and the instance label "[0-9][0-9][0-9][0-9]" select-able by character type (letters vs. numbers) it seems appropriate to preserve that ability. 

Also, the overhead of a UUID call is concerning, as it must at least interface with the MAC address, the clock, pseudo-random generator, and a hash system. Was worried a UUID call might slow down file parsing.

Open to ideas if anyone has a suggestion.

ToxicBakery: Maybe we should consider using an atomic number generator then. Maybe useful as the object IDs should be consistent as long as the model or loading order does not change.

Ie. AtomicInteger getAndIncrement()
https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/atomic/AtomicInteger.html#getAndIncrement--

contriteobserver: Implemented an `AtomicInteger` based identification mechanism, the `PosingDoll` object graph now looks like:

 I/MaterialNamingRenderer.InitScreen: object graph
    -->GROUP default
    	-->Arm.Left1002
    	-->Arm.Right1003
    	-->Leg.Left1004
    	-->Leg.Right1005
    	-->Head1006
    	-->Standing_Pose

ToxicBakery (approved):

PR #2426, opened 2022-01-20
Enhanced SpriteSheet Animation

contriteobserver:
`SpriteSheetMaterialPlugin` is now controlled via `SpriteSheetAnimation3D`, rather than directly off `SystemClock`, which enables sprite based characters to switch between animations or behaviors at runtime.

This enhancement provides Rajawali with two different capabilities for managing SpriteSheet animation.
 - `SpriteSheetMaterialPlugin` and `SpriteSheetAnimation3D` control all textures on a single `Object3D`
 - whereas `SpriteSheetAnimation` controls a single texture on all objects using that texture


ToxicBakery (approved):
Nice

Issue #2429, opened 2022-01-22
Some `ScreenQuad` materials spew error messages at frame rate


contriteobserver:
#### Rajawali Version or Branch
master

#### Summary
`EffectsPass` that directly load `VertexShader` in the `ScreenQuad` material, complain of an out of bounds vertex (normal) attribute, a workaround is to derive a class that does not call upon normal attributes.

#### Steps to Reproduce
```
    static class RadialBlurRenderer extends Renderer {
        private PostProcessingManager mPostProcessingManager;

        public RadialBlurRenderer(Context context) {
            super(context);
        }

        @Override
        protected void initScene() {
            try {
                DirectionalLight key = new DirectionalLight(-4, -4, -4);
                key.setPower(2);
                getCurrentScene().addLight(key);

                Object3D obj = new Cube(1);
                obj.setMaterial(new Material());
                obj.getMaterial().setDiffuseMethod(new DiffuseMethod.Lambert());
                obj.getMaterial().setSpecularMethod(new SpecularMethod.Phong());
                obj.getMaterial().enableLighting(true);
                obj.setColor(Color.YELLOW & Color.GRAY);
                getCurrentScene().addChild(obj);

                getCurrentCamera().setUpAxis(Vector3.Axis.Z);
                getCurrentCamera().setPosition(3,2,1);
                getCurrentCamera().setLookAt(obj.getPosition());
                getCurrentCamera().setFieldOfView(45);

                RenderPass renderPass = new RenderPass(getCurrentScene(), getCurrentCamera(), 0);
                RadialBlurPass blurPass = new RadialBlurPass();
                blurPass.setRenderToScreen(true);

                mPostProcessingManager = new PostProcessingManager(this);
                mPostProcessingManager.addPass(renderPass);
                mPostProcessingManager.addPass(blurPass);
            } catch (Exception e) {
                Log.e(getClass().getSimpleName(), e.getMessage());
            }
        }

        @Override
        public void onRender(final long elapsedTime, final double deltaTime) {
            mPostProcessingManager.render(elapsedTime, deltaTime);
        }

        @Override
        public void onOffsetsChanged(float v, float v1, float v2, float v3, int i, int i1) {
        }

        @Override
        public void onTouchEvent(MotionEvent motionEvent) {

        }

    }
```

#### Trace or Log Output
```
E/emuglGLESv2_enc: a vertex attribute index out of boundary is detected. Skipping corresponding vertex attribute. buf=0xccd03690
[...]

PR #2430, opened 2022-01-22
cleaned up `VertexShader` dependencies

contriteobserver:
addresses issue #2429

added helper functions `castFloat` and `assignDivide` to allow the shader definition to be mostly interpreted through `ShaderBase`


contriteobserver:
closing as there seems to be a better approach

